PR #1, opened 2023-10-06
Handle server side DTLS handshake

mickel8:
This PR adds support for server side DTLS handshake i.e. we always act as a server and the other side must always initiate the handshake


lib/ex_webrtc/peer_connection.ex
```diff
@@ -202,9 +237,55 @@ defmodule ExWebRTC.PeerConnection do
     {:noreply, state}
   end
 
+  @impl true
+  def handle_info({:ex_ice, _from, {:data, data}}, %{dtls_finished: false} = state) do
+    case ExDTLS.process(state.dtls_client, data) do
+      {:handshake_packets, packets} when state.ice_state in [:connected, :completed] ->
+        :ok = ICEAgent.send_data(state.ice_agent, packets)
+
```
LVala, line 245: no `{:noreply, ...}` return value in this arm of the case block.

LVala (approved):

PR #2, opened 2023-10-18
Create transceivers on set_remote_description

mickel8:
Definitely not perfect and for sure bugged but maybe this can be a base for further work.

The idea is that we iterate over mlines in SDP, we try to find a transceiver for a given mline, if it exists we potentially update its configuraiton. If it doesn't exist we create a new one. At the end we compare previous transceivers with new transceivers and try to find new MediaStreamTracks

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -7,7 +7,7 @@ defmodule ExWebRTC.PeerConnection do
 
   alias __MODULE__.Configuration
   alias ExICE.ICEAgent
-  alias ExWebRTC.{IceCandidate, SessionDescription}
+  alias ExWebRTC.{IceCandidate, MediaStreamTrack, RTPTransceiver, SessionDescription}
 
   import ExWebRTC.Utils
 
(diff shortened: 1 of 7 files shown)


lib/ex_webrtc/peer_connection.ex
```diff
@@ -293,13 +302,43 @@ defmodule ExWebRTC.PeerConnection do
...
+        # only take new transceivers that can receive tracks
+        |> Enum.filter(fn tr ->
+          RTPTransceiver.find_by_mid(state.transceivers, tr.mid) == nil and
+            tr.direction in [:recvonly, :sendrecv]
+        end)
+        |> Enum.map(fn tr -> MediaStreamTrack.from_transceiver(tr) end)
+
+      for track <- new_remote_tracks do
+        notify(state.owner, {:track, track})
+      end
```
LVala, line 323: I suppose you know that and this is only temporary, but I believe this notification should appear when only tracks are negotiated.

LVala (approved):

lib/ex_webrtc/peer_connection.ex
mickel8: Now we accept everything :D

lib/ex_webrtc/rtp_transceiver.ex
```diff
@@ -0,0 +1,38 @@
+defmodule ExWebRTC.RTPTransceiver do
```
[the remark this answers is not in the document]
LVala: My bad, I meant `RTC` prefix, no issues here.

PR #7, opened 2023-10-30
Add/refactor `set_local_description` and `create_answer`

LVala:


lib/ex_webrtc/peer_connection.ex
```diff
@@ -192,26 +192,34 @@ defmodule ExWebRTC.PeerConnection do
...
     {:ok, ice_ufrag, ice_pwd} = ICEAgent.get_local_credentials(state.ice_agent)
     {:ok, dtls_fingerprint} = ExDTLS.get_cert_fingerprint(state.dtls_client)
 
     answer = %ExSDP{ExSDP.new() | timing: %ExSDP.Timing{start_time: 0, stop_time: 0}}
 
+    answer =
+      case ExSDP.get_attribute(remote_offer, :ice_options) do
+        {:ice_options, "trickle"} = attr -> ExSDP.add_attribute(answer, attr)
+        _other -> answer
+      end
```
mickel8, line 206: Just FYI we don't support anything other than `trickle` ice so this attribute has to always be in offer/answer

lib/ex_webrtc/peer_connection.ex
```diff
@@ -392,17 +400,33 @@ defmodule ExWebRTC.PeerConnection do
...
+  defp apply_local_description(type, sdp, state) do
+    new_transceivers = update_local_transceivers(type, state.transceivers, sdp)
+    state = set_description(:local, type, sdp, state)
+
+    {:ok, %{state | transceivers: new_transceivers}}
+  end
+
+  defp update_local_transceivers(:offer, transceivers, sdp) do
+    sdp.media
+    |> Enum.zip(transceivers)
```
mickel8, line 412: I am really not sure about zipping those two lists. We can discuss this offline but consider the following scenario:
1. remote sends to us two tracks
2. at some point, remote stops one track - the first one in SDP. In such the case we should probably delete our first transceiver as it will be marked as `stopped`. We will have two mlines (the first one `inactive`, the second one `recvonly` from local perspective) and one transceiver. 
3. If at this point, we decide to add a track on local side, we will add a new transceiver that will correspond to the first mline (which will be recycled). So we will end up with two transceivers and two mlines but the first transceiver will correspond to the second mline and the second transceiver will correspond to the first mline.

Does it make sense? 

That's why I didn't decide to zip those two lists in `RTPTransceiver.update_or_create`.

We can also merge this as is and go back to the problem once it appears i.e. when we support track removal or stopping transceivers.

LVala: Applying the local offer is the moment when mids are assigned to the transceivers, so if we want to prevent the issues that you have mentioned (so basically order of transceivers != order of mlines) we would need to map mline <-> transceiver in some other way (keep the mline index in transceiver state?). 

For now I would probably leave it as it is, as that a bigger issue.

Also there's question: what if I create an offer, remove a transceiver and then apply the offer? Should it fail? Will need to take care of such cases in the near future probably.

PR #8, opened 2023-10-31
Assign mids in `create_offer`

LVala:
Quick PR with the change about MID assignment that we were discussing.
Also moved `to_mline` function to `SDPUtils` to be more consistent. When we start to handle subsequent offers/answers, it's going to be refactored anyways.

### DIFF
--- a/lib/ex_webrtc/sdp_utils.ex
+++ b/lib/ex_webrtc/sdp_utils.ex
@@ -46,7 +46,6 @@ defmodule ExWebRTC.SDPUtils do
     |> ExSDP.Media.add_attributes(attributes ++ media_formats)
   end
 
-  # TODO: refactor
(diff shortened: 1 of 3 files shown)


lib/ex_webrtc/sdp_utils.ex
```diff
@@ -46,6 +46,46 @@ defmodule ExWebRTC.SDPUtils do
     |> ExSDP.Media.add_attributes(attributes ++ media_formats)
   end
 
+  # TODO: refactor
```
mickel8, line 49: I would get rid of TODO: refactor. Let's write a code that is acceptable to a degree that it can be merged and I believe this one can be merged :stuck_out_tongue:

mickel8 (approved):

PR #10, opened 2023-11-03
Move DTLS utilities to `DTLSTransport` module

LVala:
Additionally, fixes some bugs with active mode.

### DIFF
--- a/lib/ex_webrtc/dtls_transport.ex
+++ b/lib/ex_webrtc/dtls_transport.ex
@@ -46,7 +46,7 @@ defmodule ExWebRTC.DTLSTransport do
     %__MODULE__{dtls | client: client}
   end
 
-  def start(%{ice_state: ice_state} = dtls, :active) do
+  def start(dtls, :active) do
     {:ok, client} =
       ExDTLS.start_link(
         client_mode: true,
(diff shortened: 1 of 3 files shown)


lib/ex_webrtc/dtls_transport.ex
```diff
@@ -0,0 +1,128 @@
...
+    :ice_agent,
+    :ice_state,
+    :client,
+    :buffered_packets,
+    :cert,
+    :pkey,
+    :fingerprint,
+    :mode,
+    finished: false,
+    should_start: false
```
mickel8, line 18: I would favor using `mode` field instead, which currerntly remains unset

lib/ex_webrtc/dtls_transport.ex
```diff
@@ -0,0 +1,128 @@
...
+        :ok = ICEAgent.send_data(dtls.ice_agent, packets)
+        %__MODULE__{dtls | should_start: false}
+      else
+        dtls
+      end
+
+    dtls =
+      if dtls.buffered_packets do
+        Logger.debug("Sending buffered DTLS packets")
+        ICEAgent.send_data(dtls.ice_agent, dtls.buffered_packets)
```
mickel8, line 75: I assume this was coppied but could we always match against `:ok`?

mickel8 (approved):

PR #12, opened 2023-11-06
Setup SRTP decryption

LVala:


lib/ex_webrtc/dtls_transport.ex
```diff
@@ -93,31 +97,67 @@ defmodule ExWebRTC.DTLSTransport do
...
+        {:ok, dtls}
     end
   end
+
+  def process_data(dtls, <<f, _::binary>> = data) when f in 128..191 and dtls.finished do
+    case ExLibSRTP.unprotect(dtls.srtp, data) do
+      {:ok, payload} ->
+        {:ok, dtls, payload}
+
+      {:error, reason} = err when reason in [:reply_old, :reply_fail] ->
```
mickel8, line 135: Let's ignore all errors for now. Maybe just log warning so we see them

mickel8 (approved):

PR #13, opened 2023-11-06
Configure codecov

mickel8:
* disabled github notifications that something is not covered by tests
* we don't need to wait for other checks to run codecov
* changed the layout of codecov comment so we see more info. Behavior `default` means that the comment is updated when new commits appear
* disabled coverage threshold


LVala (approved):
I was about to do similar changes, but it seems that this time I won the "be very annoyed and do nothing about it" contest.

PR #14, opened 2023-11-06
Make codecs and rtp header extensions configurable

mickel8:


lib/ex_webrtc/peer_connection/configuration.ex
```diff
@@ -12,32 +66,157 @@ defmodule ExWebRTC.PeerConnection.Configuration do
...
+    struct!(__MODULE__, options)
+  end
+
+  @doc false
+  @spec is_supported_codec(t(), RTPCodecParameters.t()) :: boolean()
+  def is_supported_codec(config, codec) do
+    # This function doesn't check if rtcp-fb is supported.
+    # Instead, `is_supported_rtcp_fb` has to be used to filter out
+    # rtcp-fb that are not supported.
+    Enum.find(
```
LVala, line 145: `Enum.find` doesn't return a `boolean`, contrary to what the function spec suggests.

test/configuration_test.exs
LVala, line 1: All is good, but this is more like an integration tests, shouldn't there be some more function specific tests, I mean functions from the `Configuration` module?

LVala (approved):

PR #18, opened 2023-11-21
Convert `RTPTransceiver` to separate process

LVala:


LVala:
For the time being, `RTPTransceiver` handling will be left in the `PeerConnection` process, we'll take care of parallelizing it later on.

PR #21, opened 2023-11-24
Allow for sending data

LVala:
Additionally, modify the examples to create a media loop (browser -> Elixir PeerConnection -> browser).


lib/ex_webrtc/dtls_transport.ex
```diff
@@ -121,18 +127,25 @@ defmodule ExWebRTC.DTLSTransport do
...
     {:noreply, state}
   end
 
   @impl true
-  def handle_cast({:send_data, _data}, state) do
-    Logger.error(
-      "Attempted to send data when DTLS handshake was not finished or ICE Transport is unavailable"
-    )
+  def handle_cast({:send_rtp, _data}, state) do
+    Logger.warning("Attemped to send RTP before DTLS handshake has been finished")
```
mickel8, line 142: Let's explicitly state that we ignore the packet. This way we will always be sure reading docs that this behaviour doesn't have any impact on the PC.

```suggestion
    Logger.warning("Attemped to send RTP before DTLS handshake has been finished. Ignoring.")
```

lib/ex_webrtc/dtls_transport.ex
```diff
@@ -195,16 +209,16 @@ defmodule ExWebRTC.DTLSTransport do
         state = %{state | buffered_packets: packets}
         update_dtls_state(state, :connecting)
 
-      {:handshake_finished, _, remote_keying_material, profile, packets} ->
+      {:handshake_finished, lkm, rkm, profile, packets} ->
         Logger.debug("DTLS handshake finished")
         ICEAgent.send_data(state.ice_agent, packets)
         # TODO: validate fingerprint
-        state = setup_srtp(state, remote_keying_material, profile)
+        state = setup_srtp(state, lkm, rkm, profile)
```
mickel8, line 216: we could make setup_srtp return :ok

lib/ex_webrtc/peer_connection.ex
```diff
@@ -326,26 +337,56 @@ defmodule ExWebRTC.PeerConnection do
...
+  def handle_cast({:send_rtp, track_id, packet}, state) do
+    sdes_id =
+      Enum.find_value(state.demuxer.extensions, fn
+        {ext_id, {ExRTP.Packet.Extension.SourceDescription, :mid}} -> ext_id
+        _ -> nil
+      end)
+
+    # TODO: iterating over transceivers is not optimal
+    # but this is, most likely, going to be refactored anyways
+    maybe_transceiver =
```
mickel8, line 364:
```suggestion
    transceiver =
```

lib/ex_webrtc/peer_connection.ex
```diff
@@ -377,11 +418,16 @@ defmodule ExWebRTC.PeerConnection do
...
 
   @impl true
-  def handle_info({:dtls_transport, _pid, {:rtp_data, data}}, state) do
-    case Demuxer.demux(state.demuxer, data) do
-      {:ok, demuxer, mid, packet} ->
-        notify(state.owner, {:data, {mid, packet}})
-        {:noreply, %__MODULE__{state | demuxer: demuxer}}
+  def handle_info({:dtls_transport, _pid, {:rtp, data}}, state) do
+    with {:ok, demuxer, mid, packet} <- Demuxer.demux(state.demuxer, data),
+         %RTPTransceiver{} = t <- Enum.find(state.transceivers, &(&1.mid == mid)) do
```
mickel8, line 423: Not sure about this one. I belive this should be in the `do end` body as this should always be true

lib/ex_webrtc/dtls_transport.ex
LVala: why tho? Do you mean that `setup_srtp` could fail when provided with these keying materials or profile?

lib/ex_webrtc/peer_connection.ex
LVala: Yeah, but atm demuxer does not check that, I could change this in this pull request, but the whole receiving/sending data business is quite temporary anyways, isn't it?

PR #25, opened 2023-11-27
Add ICETransport behaviour

mickel8:


lib/ex_webrtc/dtls_transport.ex
```diff
@@ -60,19 +67,20 @@ defmodule ExWebRTC.DTLSTransport do
...
 
-    {:ok, ice_agent} = ice_module.start_link(:controlled, ice_config)
+    {:ok, ice_transport_state} = ice_transport.start_link(:controlled, ice_config)
     srtp = ExLibSRTP.new()
 
     state = %{
       owner: owner,
-      ice_agent: ice_agent,
+      ice_transport: ice_transport,
+      ice_transport_state: ice_transport_state,
```
LVala, line 83: `ice_transport_state` is very misleading, as it has nothing to do with any state.
IMO the `ice_transport` should be called maybe like `ice_transport_impl` or `ice_transport_module` and the `ice_transport_state` should be `ice_transport`.

Or maybe `ice_transport` and `ice_pid`, like in the `PeerConnection`.

test/ex_webrtc/dtls_transport_test.exs
```diff
@@ -36,21 +52,20 @@ defmodule ExWebRTC.DTLSTransportTest do
...
-    %{dtls: dtls, ice: ice}
+    %{dtls: dtls, ice_transport: ice_transport, ice_pid: ice_pid}
   end
 
-  test "forwards non-data ICE messages", %{ice: ice} do
+  test "forwards non-data ICE messages", %{ice_transport: ice_transport, ice_pid: ice_pid} do
     message = :connected
 
-    FakeICEAgent.send_dtls(ice, message)
+    ice_transport.send_dtls(ice_pid, message)
```
LVala, line 65: Well, it's not like you're going to use anything else other than `MockICETransport` as an `ICETransport`, but you went all the way through with the mocking game, I guess.

LVala (approved):

test/ex_webrtc/dtls_transport_test.exs
mickel8: Not really, tbh it's my first time playing with mocks but I did it to be able to easily change MockICETransport to something other or to rename MockICETransport

PR #27, opened 2023-11-28
Move ICETransport initialization to PeerConnection

mickel8:


lib/ex_webrtc/dtls_transport.ex
```diff
@@ -30,21 +30,21 @@ defmodule ExWebRTC.DTLSTransport do
   @type dtls_state() :: :new | :connecting | :connected | :closed | :failed
 
   @doc false
-  @spec start_link(ICETransport.t(), Keyword.t()) :: GenServer.on_start()
-  def start_link(ice_transport \\ DefaultICETransport, ice_config) do
+  @spec start_link(ICETransport.t(), pid()) :: GenServer.on_start()
```
LVala, line 33: Using `pid()` instead of `GenServer.server()` as a spec here is alright, but it makes it a little bit inconsistent with stuff in `DTLSTransport`, for instance.

LVala (approved):

PR #31, opened 2023-12-04
 Negotiation improvements

LVala:


lib/ex_webrtc/peer_connection.ex
```diff
@@ -470,91 +458,138 @@ defmodule ExWebRTC.PeerConnection do
...
          :ok <- SDPUtils.ensure_bundle(sdp),
          {:ok, {ice_ufrag, ice_pwd}} <- SDPUtils.get_ice_credentials(sdp),
          {:ok, {:fingerprint, {:sha256, peer_fingerprint}}} <- SDPUtils.get_cert_fingerprint(sdp),
-         {:ok, new_transceivers} <-
-           update_remote_transceivers(state.transceivers, sdp, state.config) do
-      :ok =
-        state.ice_transport.set_remote_credentials(state.ice_pid, ice_ufrag, ice_pwd)
+         {:ok, dtls_role} <- SDPUtils.dtls_role_from_remote(sdp),
+         {:ok, transceivers} <- update_transceivers(sdp, state) do
+      state = %{state | can_trickle_ice_candidates: SDPUtils.check_if_trickle(sdp)}
```
mickel8, line 505: Avoid adding fields we are currently not using. We will add them once they are needed

mickel8 (changes requested):

lib/ex_webrtc/peer_connection.ex
```diff
@@ -470,91 +458,138 @@ defmodule ExWebRTC.PeerConnection do
...
+  defp next_state(:have_remote_offer, _, _), do: {:error, :invalid_state}
+  defp next_state(:have_local_pranswer, :local, :pranswer), do: {:ok, :have_local_pranswer}
+  defp next_state(:have_local_pranswer, :local, :answer), do: {:ok, :stable}
+  defp next_state(:have_local_pranswer, _, _), do: {:error, :invalid_state}
+  defp next_state(:have_remote_pranswer, :remote, :pranswer), do: {:ok, :have_remote_pranswer}
+  defp next_state(:have_remote_pranswer, :remote, :answer), do: {:ok, :stable}
+  defp next_state(:have_remote_pranswer, _, _), do: {:error, :invalid_state}
+
+  defp check_altered(:offer, sdp, %{last_offer: sdp}), do: :ok
+  defp check_altered(type, sdp, %{last_answer: sdp}) when type in [:answer, :pranswer], do: :ok
```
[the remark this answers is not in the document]
LVala: I make the `set_X_description` return `:not_implemented` when called with `:pranswer`, to this cannot lead to errors. I think I would rather just leave it, this much of future -proofing won't hurt.

lib/ex_webrtc/peer_connection.ex
```diff
@@ -470,91 +458,138 @@ defmodule ExWebRTC.PeerConnection do
...
-    end
+      pt_to_mid = SDPUtils.get_payload_to_mid(sdp)
+      demuxer = %Demuxer{state.demuxer | mid_ext_id: mid_ext_id, pt_to_mid: pt_to_mid}
+      state = %{state | demuxer: demuxer}
 
-    {:ok, %{state | transceivers: new_transceivers, demuxer: demuxer}}
-  end
+      if state.ice_gathering_state == :new do
+        state.ice_transport.gather_candidates(state.ice_pid)
+      end
```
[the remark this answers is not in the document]
LVala: Why we would need remote ICE credentials to start gathering candidates locally?
A very typical flow without trickle would look something like this:
createOffer -> setLocalDescription -> wait for candidates to gather -> getLocalDescription -> send the description to remote peer

We need to start gathering right away.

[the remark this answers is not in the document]
mickel8: For future reference:
* we don't need remote password to calculate message-integrity
* we might need remote ufrag to check whether username is correct
* even if we need remote ufrag and we don't have it (we decline conn check), if we execute conn check on the same pair on our side, the remote party will re-sechudle its conn check
* pion checks whole username 
* RFC 8445 suggests that username is valid when it consists fo two parts separtaed with a colon when the first par is our (the side that receives conn check) ufrag

lib/ex_webrtc/sdp_utils.ex
```diff
@@ -81,6 +81,16 @@ defmodule ExWebRTC.SDPUtils do
     end
   end
 
+  @spec ensure_rtcp_mux(ExSDP.t()) :: :ok | {:error, :rtcp_mux_not_indicated}
+  def ensure_rtcp_mux(sdp) do
+    sdp.media
+    |> Enum.all?(&(ExSDP.Media.get_attribute(&1, :rtcp_mux) == :rtcp_mux))
+    |> case do
+      true -> :ok
+      false -> {:error, :rtcp_mux_not_included}
```
mickel8, line 90: To be consistent with other functions 

```suggestion
      false -> {:error, :missing_rtcp_mux}
```

lib/ex_webrtc/peer_connection.ex
```diff
@@ -458,61 +457,59 @@ defmodule ExWebRTC.PeerConnection do
...
-
+         {:ok, dtls_role} <- SDPUtils.get_dtls_role(sdp),
+         {:ok, transceivers} <- update_transceivers(state, sdp) do
       :ok = state.ice_transport.set_remote_credentials(state.ice_pid, ice_ufrag, ice_pwd)
 
       for candidate <- SDPUtils.get_ice_candidates(sdp) do
         state.ice_transport.add_remote_candidate(state.ice_pid, candidate)
       end
 
+      # infer our role from the remote role
```
mickel8, line 499: Appreciate the comment

lib/ex_webrtc/peer_connection.ex
```diff
@@ -470,91 +457,146 @@ defmodule ExWebRTC.PeerConnection do
...
-        setup =
-          case setup do
-            :active -> :passive
-            :passive -> :active
-          end
+      state = %{state | demuxer: Demuxer.update(state.demuxer, sdp)}
 
-        DTLSTransport.start_dtls(state.dtls_transport, setup, peer_fingerprint)
-      end
+      # TODO: for now, we emit track event for every new transceiver
```
mickel8, line 505: For now? Is this subject to change?

LVala: Renegotation can result in a new track on existing transceiver and we don't handle that. That's what I meant.

lib/ex_webrtc/sdp_utils.ex
```diff
@@ -140,10 +140,15 @@ defmodule ExWebRTC.SDPUtils do
     |> Enum.map(fn {"candidate", attr} -> attr end)
   end
 
-  @spec get_dtls_role(ExSDP.t()) :: {:ok, :active | :passive} | {:error, :not_dtls_role}
+  @spec get_dtls_role(ExSDP.t()) ::
+          {:ok, :active | :passive | :actpass} | {:error, :not_dtls_role}
```
mickel8, line 144:
```suggestion
          {:ok, :active | :passive | :actpass} | {:error, :missing_dtls_role}
```

lib/ex_webrtc/peer_connection.ex
```diff
@@ -488,7 +484,7 @@ defmodule ExWebRTC.PeerConnection do
          :ok <- SDPUtils.ensure_rtcp_mux(sdp),
          {:ok, {ice_ufrag, ice_pwd}} <- SDPUtils.get_ice_credentials(sdp),
          {:ok, {:fingerprint, {:sha256, peer_fingerprint}}} <- SDPUtils.get_cert_fingerprint(sdp),
-         {:ok, dtls_role} <- SDPUtils.get_dtls_role(sdp),
+         {:ok, setup} <- SDPUtils.get_dtls_role(sdp),
```
mickel8, line 487: Why changing this? `dtls_role` was consistent with `get_dtls_role` name

lib/ex_webrtc/peer_connection.ex
mickel8: Could we write this in the comment?

mickel8 (approved):

PR #34, opened 2023-12-08
Properly handle new remote tracks

LVala:


test/ex_webrtc/peer_connection_test.exs
```diff
@@ -37,6 +37,9 @@ defmodule ExWebRTC.PeerConnectionTest do
 
     assert_receive {:ex_webrtc, ^pc, {:track, %MediaStreamTrack{kind: :audio}}}
 
+    {:ok, answer} = PeerConnection.create_answer(pc)
```
mickel8, line 40: Would be nice to add a test that should fail in the previous implementation but should pass in the current one

lib/ex_webrtc/peer_connection.ex
```diff
@@ -619,19 +623,71 @@ defmodule ExWebRTC.PeerConnection do
...
-    Enum.reduce_while(sdp.media, {:ok, transceivers}, fn mline, {:ok, transceivers} ->
-      case ExSDP.Media.get_attribute(mline, :mid) do
-        {:mid, mid} ->
-          transceivers = RTPTransceiver.update_or_create(transceivers, mid, mline, config)
-          {:cont, {:ok, transceivers}}
+  defp update_transceivers(state, sdp) do
+    Enum.reduce(sdp.media, state.transceivers, fn mline, transceivers ->
+      {:mid, mid} = ExSDP.Media.get_attribute(mline, :mid)
+
+      # TODO: reusing transceivers
```
mickel8, line 630: Could we elaborate on this comment? I'm slowly getting lost

mickel8 (changes requested):

lib/ex_webrtc/peer_connection.ex
LVala: I meant "recycling transceivers", we don't do that at the moment.

lib/ex_webrtc/peer_connection.ex
```diff
@@ -619,19 +623,71 @@ defmodule ExWebRTC.PeerConnection do
...
+      {:mid, mid} = ExSDP.Media.get_attribute(mline, :mid)
+      {idx, transceiver} = find_transceiver(transceivers, mid)
+
+      direction = SDPUtils.get_media_direction(mline)
+
+      direction =
+        case source do
+          :local -> direction
+          :remote -> reverse_direction(direction)
+        end
```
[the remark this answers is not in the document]
LVala: As we discussed offline, this implementation is actually true to the spec.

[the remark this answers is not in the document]
mickel8: For future reference, see this comment https://github.com/w3c/webrtc-pc/issues/2751#issuecomment-1851583208

mickel8 (approved):

PR #44, opened 2023-12-20
Add IVF writer

mickel8:


lib/ex_webrtc/media/ivf_writer.ex
```diff
@@ -0,0 +1,88 @@
...
+
+  defguard update_header?(writer)
+           when writer.frames_cnt >= writer.update_header_after and
+                  rem(writer.frames_cnt, writer.update_header_after) == 0
+
+  @doc """
+  Creates a new IVF writer.
+
+  Initially, IVF header is written with `num_frames` and 
+  is updated every `num_frames` by `num_frames`.
```
LVala, line 24: I wonder if it wouldn't be better to explicitly update the header after all the frames are written? Updating every x frames seems kind error-prone, what if no frames are available, but the header was not updated? Updating header every new frame also seems wastefuf.

I'm not familiar with IVF, but that's just  an observation.

LVala (approved):

lib/ex_webrtc/media/ivf_writer.ex
mickel8: I can add `close` function that I had before that will update header when called. However, it looks like there are some cases where we are not able to call `close` e.g. sending SIGINT (CTRL+C) when executing Elixir script.

I think this should be okay to have a header indicating more frames than there are in the file. Also, Pion starts with I 900 frames and updates header on calling `close`.

lib/ex_webrtc/media/ivf_reader.ex
```diff
@@ -84,37 +84,34 @@ defmodule ExWebRTC.Media.IVFReader do
...
-           num_frames: num_frames,
-           unused: unused
-         }}
-
-      _other ->
-        {:error, :invalid_file}
+  @spec open(Path.t()) :: {:ok, IVFHeader.t(), t()} | {:error, term()}
+  def open(path) do
+    with {:ok, file} <- File.open(path),
+         <<"DKIF", 0::little-16, 32::little-16, fourcc::little-32, width::little-16,
```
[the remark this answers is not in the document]
mickel8: I would keep this to be consistent

PR #46, opened 2023-12-20
Add save_to_file example

mickel8:


examples/save_to_file/example.exs
```diff
@@ -0,0 +1,206 @@
...
+  defp handle_webrtc_message({:track, _track}, state) do
+    <<fourcc::little-32>> = "VP80"
+
+    # Width, height and FPS (timebase_denum/num)
+    # are the same as we set in video constraints
+    # on the frontend side (in getUserMedia).
+    # However, keep in mind they can change in time
+    # so this is best effort saving.
+    # num_frames is set to 900 but it will be updated 
+    # on calling IVFWriter.close
```
LVala, line 159: There's no call to `IVFWriter.close` anywhere here, I was scratching my head thinking I missed it in the `IFVWriter` PR review, but there seems to be no such function.

mickel8: Yeah, sorry there was such function but I deleted it and restored again

PR #49, opened 2024-01-01
Add removeTrack

mickel8:


test/ex_webrtc/peer_connection_test.exs
```diff
@@ -36,79 +36,83 @@ defmodule ExWebRTC.PeerConnectionTest do
                  fingerprint: {:sha256, @fingerprint}
                )
 
-  test "negotiation needed" do
```
mickel8, line 39: I had to separate those cases as it's hard to determine which one exactly was failing

test/ex_webrtc/peer_connection_test.exs
```diff
@@ -36,79 +36,83 @@ defmodule ExWebRTC.PeerConnectionTest do
...
+      {:ok, pc} = PeerConnection.start_link()
+      track = MediaStreamTrack.new(:video)
+      {:ok, _tr} = PeerConnection.add_track(pc, track)
+      assert_receive {:ex_webrtc, ^pc, :negotiation_needed}
+    end
 
-    {:ok, answer} = PeerConnection.create_answer(pc2)
-    :ok = PeerConnection.set_local_description(pc2, answer)
-    :ok = PeerConnection.set_remote_description(pc, answer)
+    test "is fired on remove_track" do
```
mickel8, line 60: A new one

lib/ex_webrtc/peer_connection.ex
```diff
@@ -690,28 +772,35 @@ defmodule ExWebRTC.PeerConnection do
...
-    # because current_direction is set in set_remote_description
+  # see W3C WebRTC 5.1.1
+  defp process_remote_track(transceiver, mline, owner) do
     direction = SDPUtils.get_media_direction(mline)
 
-    if direction in [:sendrecv, :sendonly] and
-         tr.current_direction not in [:sendrecv, :recvonly] do
-      notify(owner, {:track, tr.receiver.track})
+    cond do
+      direction in [:sendrecv, :recvonly] and
```
mickel8, line 791: I changed from `[:sendrecv, :sendonly]` to `[:sendrecv, :recvonly]` according to spec

LVala: Remember that the W3C 5.1.1 assumes that `direction` is the remote m-line direction from this peer's point of view (it is swapped in 4.4.1.5.4.7.10.2.10, "10. Let `direction` be ...). We don't do the swap (afaik), that's why I used the `:sendonly`. If I understand correctly, now your trying to emit the track event for remote recvonly tracks (so sendonly from our peer's perspective).

test/ex_webrtc/peer_connection_test.exs
```diff
@@ -50,12 +50,34 @@ defmodule ExWebRTC.PeerConnectionTest do
...
+    {:ok, pc} = PeerConnection.start_link()
+    {:ok, pc2} = PeerConnection.start_link()
+    {:ok, tr} = PeerConnection.add_transceiver(pc, :audio)
+    assert_receive {:ex_webrtc, ^pc, :negotiation_needed}
+    :ok = negotiate(pc, pc2)
+    :ok = PeerConnection.set_transceiver_direction(pc, tr.id, :sendonly)
+    refute_receive {:ex_webrtc, ^pc, :negotiation_needed}
+    :ok = PeerConnection.set_transceiver_direction(pc, tr.id, :recvonly)
+    assert_receive {:ex_webrtc, ^pc, :negotiation_needed}
+
```
LVala, line 63: This test starts to get a bit big. Maybe split into separate tests under "negotiation needed" describe?

mickel8: Yeah, I am going to refactor this but in a separate PR

lib/ex_webrtc/peer_connection.ex
mickel8: Good point, I didn't check this

lib/ex_webrtc/peer_connection.ex
```diff
@@ -522,6 +585,24 @@ defmodule ExWebRTC.PeerConnection do
...
+      # This should result in emitting track mute event.
+      if type == :answer do
+        for mline <- sdp.media do
+          {:mid, mid} = ExSDP.get_attribute(mline, :mid)
+          tr = Enum.find(state.transceivers, fn tr -> tr.mid == mid end)
+          direction = SDPUtils.get_media_direction(mline)
+
+          if direction in [:sendonly, :inactive] and tr.fired_direction in [:sendrecv, :recvonly] do
+            notify(state.owner, {:track_muted, tr.receiver.track.id})
+          end
```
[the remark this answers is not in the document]
mickel8: track_muted is also emitted when the other side has removed a track (called removeTrack)

fired_direction is also needed to not emit multiple track events when applying remote offer multiple times in a row. See https://github.com/elixir-webrtc/ex_webrtc/blob/2ed431de96a098ab5ba119723d4887c69281148d/test/ex_webrtc/peer_connection_test.exs#L152-L154

> although implementing this very specific feature (muted event is emitted ONLY when user sets transceiver's direction to inactive between sRD and createAnswer?)

Btw. that's how the remote incoming track can be rejected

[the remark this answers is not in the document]
mickel8: And I could resolve this TODO https://github.com/elixir-webrtc/ex_webrtc/blob/master/lib/ex_webrtc/peer_connection.ex#L706-L708

lib/ex_webrtc/peer_connection.ex
```diff
@@ -522,6 +585,24 @@ defmodule ExWebRTC.PeerConnection do
         state.ice_transport.gather_candidates(state.ice_pid)
       end
 
+      # See W3C WebRTC 4.4.1.5 when "description"
```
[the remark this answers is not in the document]
mickel8: I don't like using them as they can't be linked, are subject to change and it depends on whether we take a look at current or future version

LVala (approved):

PR #50, opened 2024-01-03
Ogg writer

LVala:
Adds the `OggWriter` and its tests.

Example will be updated in the following Pull Request.

### DIFF
--- a/examples/send_from_file/example.exs
+++ b/examples/send_from_file/example.exs
@@ -161,7 +161,7 @@ defmodule Peer do
         Process.send_after(self(), :send_audio_packet, duration)
         # values set to 0 are handled by PeerConnection.set_rtp
         rtp_packet = OpusPayloader.payload(packet)
-        rtp_packet = %{rtp_packet | timestamp: state.last_audio_timestamp}
+        rtp_packet = %{rtp_packet | timestamp: trunc(state.last_audio_timestamp)}
         PeerConnection.send_rtp(state.peer_connection, state.audio_track_id, rtp_packet)
 
         # OggReader.next_packet/1 returns duration in ms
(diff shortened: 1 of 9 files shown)


examples/send_from_file/example.exs
```diff
@@ -161,7 +161,7 @@ defmodule Peer do
         Process.send_after(self(), :send_audio_packet, duration)
         # values set to 0 are handled by PeerConnection.set_rtp
         rtp_packet = OpusPayloader.payload(packet)
-        rtp_packet = %{rtp_packet | timestamp: state.last_audio_timestamp}
+        rtp_packet = %{rtp_packet | timestamp: trunc(state.last_audio_timestamp)}
```
mickel8, line 164: Can this ever be non-integer? We calculate last_audio_timestamp as:  `last_audio_timestamp: state.last_audio_timestamp + timestamp_delta` where timestamp_delta is `timestamp_delta = trunc(duration * 48_000 / 1000)` (so integer)  and initial last_audio_timestamp is also integer

lib/ex_webrtc/media/ogg_reader.ex
```diff
@@ -1,6 +1,6 @@
 defmodule ExWebRTC.Media.OggReader do
```
mickel8, line 1: If we have Ogg.Header and Ogg.Packet, it might be resonable to make everything the same in this PR (including IVF modules)

mickel8 (approved):
Awesome! :tada:

lib/ex_webrtc/media/ogg_reader.ex
LVala: I was wondering about the same, but all of the Ogg submodules are `@moduledoc false` so from API'a standpoint, it's still consistent.

examples/send_from_file/example.exs
LVala: Yeas, this is just temporary (duration from `OggReader.next_packet` is now a float), I'll make there's no unnecessary truncs in the examples PR.

PR #51, opened 2024-01-04
Add replace_track

mickel8:


test/ex_webrtc/peer_connection_test.exs
```diff
@@ -544,22 +581,46 @@ defmodule ExWebRTC.PeerConnectionTest do
...
-    {:ok, _sender} = PeerConnection.add_track(pc2, track2)
+      [tr] = PeerConnection.get_transceivers(pc2)
+      :ok = PeerConnection.set_transceiver_direction(pc2, tr.id, :sendrecv)
+      :ok = PeerConnection.replace_track(pc2, tr.sender.id, track2)
 
-    :ok = negotiate(pc2, pc1)
+      {:ok, answer} = PeerConnection.create_answer(pc2)
+      :ok = PeerConnection.set_local_description(pc2, answer)
+      :ok = PeerConnection.set_remote_description(pc1, answer)
+    end
```
LVala, line 604: This does not look like it tests "send data in both (...) using one negotiation" to me.

mickel8:
@LVala as you noticed, it looks like all known use-cases might be achieved using addTrack but because I already implemented this and to be consistent with the official W3C standard I would merge. If we find out this to be problematic in maintenance, we can remove it in the future.

test/ex_webrtc/peer_connection_test.exs
mickel8: Fixed

LVala (approved):

PR #60, opened 2024-01-16
Handle ExDTLS errors

mickel8:
When we notice ExDTLS error, DTLSTransport will stop itself and because it is linked with PeerConnection and PeerConnection is linked with ICETransport, everything will stop too

### DIFF
--- a/examples/save_to_file/example.exs
+++ b/examples/save_to_file/example.exs
@@ -23,8 +23,8 @@ defmodule Peer do
     %{urls: "stun:stun.l.google.com:19302"}
   ]
 
-  def start_link() do
-    GenServer.start_link(__MODULE__, nil)
+  def start() do
+    GenServer.start(__MODULE__, nil)
   end
 
   @impl true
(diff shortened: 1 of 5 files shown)


examples/save_to_file/example.exs
```diff
@@ -217,12 +217,12 @@ defmodule Peer do
...
 
-{:ok, pid} = Peer.start_link()
+{:ok, pid} = Peer.start()
 ref = Process.monitor(pid)
 
 receive do
-  {:DOWN, ^ref, _, _, _} ->
-    Logger.info("Peer process closed. Exiting")
+  {:DOWN, ^ref, _, _, reason} ->
+    Logger.info("Peer process closed, reason: #{inspect(reason)}. Exiting")
```
mickel8, line 225: I wonder if we should rename `Peer` module to `Client`. Right now we might get the following log message:

```
21:26:32.659 [info] Received unexpected message: %{"type" => "peer_left"}

21:26:32.665 [info] Peer process closed, reason: {:shutdown, :peer_closed_for_writing}. Exiting
```

which might be misleading as we don't know WHICH peer closed for writing :thinking: 
On the other hand, I am not sure if `Client` fits well :thinking:

LVala: I don't think it's that big of a deal. Anyone running the example will most likely look into the its code and see where this messages is logged, so I don't think that's gonna be too confusing.

LVala (approved):

PR #61, opened 2024-01-17
Fix no supported codecs scenario

mickel8:
Following Firefox behaviour, when there are no supported codecs in the offer, we will end up with empty transceivers lists on both sides after finishing negotiation.

See https://github.com/w3c/webrtc-pc/issues/2927


.credo.exs
```diff
@@ -104,7 +104,7 @@
           {Credo.Check.Readability.ParenthesesInCondition, []},
           {Credo.Check.Readability.ParenthesesOnZeroArityDefs, []},
           {Credo.Check.Readability.PipeIntoAnonymousFunctions, []},
-          {Credo.Check.Readability.PredicateFunctionNames, []},
+          {Credo.Check.Readability.PredicateFunctionNames, [exit_status: 0]},
```
mickel8, line 107: TBH I don't know what we want to do with this but personally I don't think functions like `is_supported` `is_mandatory` etc are bad. Credo suggests using `supported?` and `mandatory?` which is imo less readable. See https://hexdocs.pm/credo/Credo.Check.Readability.PredicateFunctionNames.html

lib/ex_webrtc/rtp_transceiver.ex
```diff
@@ -122,27 +122,30 @@ defmodule ExWebRTC.RTPTransceiver do
...
 
       transceiver.stopping == true or transceiver.stopped == true ->
+        opts = Keyword.put(opts, :direction, :inactive)
         mline = to_mline(transceiver, opts)
         %ExSDP.Media{mline | port: 0}
 
       true ->
+        offered_direction = SDPUtils.get_media_direction(mline)
+        direction = get_direction(offered_direction, transceiver.direction)
+        opts = Keyword.put(opts, :direction, direction)
```
LVala, line 148: Was this previously unhandled? How did we handle the `sRD(offer)` -> `tr.direction=inactive` -> `sLD(answer)` scenario then?

LVala (approved):

PR #66, opened 2024-02-02
Simplify examples

LVala:
In this PR all of the examples have been rewritten to:
- not depend on `ex_ice/signalling_server`,
- be fully-fledged `mix` projects,
- make the WebRTC-specific code simpler


mickel8 (approved):
Brilliant! :gem:

PR #72, opened 2024-02-17
Allow `MediaStreamTrack.t()` as a parameter to `send_rtp/3` 

srcrip:
hey, in my experimentation, this handle_cast for `:send_rtp` always receives a full track, not an id. This change makes everything work, but if you want it to be different feel free to edit my pr.

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -725,7 +725,7 @@ defmodule ExWebRTC.PeerConnection do
   end
 
   @impl true
-  def handle_cast({:send_rtp, track_id, packet}, state) when is_integer(track_id) do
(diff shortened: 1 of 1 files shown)


srcrip:
Actually now that I'm looking at my code closer I realized it was in fact my fault. I was sending the track explicitly.

Since that's an easy mistake to make, would it be ok if I edit this function to accept a track or track id and just handle both appropriately?

srcrip:
ok I've switched things up a bit:
- added a `@doc` for `send_rtp`
- made it able to accept track_id or mediastreamtrack, that way someone can mess up the way I did and it will just work

LVala:
Thank's for the PR! SMGT, although the non-W3C functions in the API are quite experimental and may change in future versions, just keep that in mind.

We'll probably get back to it and merge on Monday.

PR #35, opened 2023-12-11
AV1 payloader

mickel8:
This PR unfortunately doesn't work and I ran out of ideas on how to push it further. Current status:
* chromium successfully decodes and displays the first video frame
* I can't see any logs indicating we are doing something wrong. I ran dev build of chromium. The only thing I can see is
```
[2098991:3:1211/162436.394069:WARNING:dav1d_decoder.cc(135)] Dav1dDecoder::Decode decoding failed with error code -11
[2098991:3:1211/162436.394213:WARNING:generic_decoder.cc(302)] Failed to decode frame with timestamp 2618223154, error code: -1
```
According to dav1d IIRC this might not be the error but I think it shouldn't be logged
image
* I assumed that sending every OBU in a separate RTP packet is ok
* I assumed that every frame is a separate Temporal Unit (in the test file this should be true if only the code checking OBU type is correct) and therefore I set marker of the last RTP packet for every IVF frame
* I tried batching OBUs e.g. by 2 or setting W flag to 1 (instead of 0 and using LEB size) - didn't help
* I tried prepending every OBU with sequence header - this gives a massive number of decoder errors (-11)
* Pion `send_from_file` example also doesn't work but here I am getting 
```
[2077476:12:1211/123520.805336:WARNING:video_rtp_depacketizer_av1.cc(314)] Mismatch in obu_size. signaled: 0, actual: 591
[2077476:12:1211/123520.838810:WARNING:video_rtp_depacketizer_av1.cc(314)] Mismatch in obu_size. signaled: 0, actual: 3
[2077476:12:1211/123520.871969:WARNING:video_rtp_depacketizer_av1.cc(314)] Mismatch in obu_size. signaled: 0, actual: 33
[2077476:12:1211/123520.904092:WARNING:video_rtp_depacketizer_av1.cc(314)] Mismatch in obu_size. signaled: 0, actual: 3
[2077476:12:1211/123520.936935:WARNING:video_rtp_depacketizer_av1.cc(314)] Mismatch in obu_size. signaled: 0, actual: 62
[2077476:12:1211/123520.970613:WARNING:video_rtp_depacketizer_av1.cc(314)] Mismatch in obu_size. signaled: 0, actual: 3
[2077476:12:1211/123520.996106:WARNING:video_receive_stream2.cc(825)] No decodable frame in 230195 us, requesting keyframe.
[...]

Issue #28, opened 2023-11-29
Roadmap


mickel8:
# 0.1.0 (01.2024)
### Media reading/writing:
* [x] IVF reader/writer - #29 #44
* [x] Ogg reader/writer - #43 #50
* [x] VP8 RTP packetizer/depacketizer - #30 #45
* [x] Opus RTP packetize/depacketizer - #47 #53
* [x] Sending from a file (without RTCP) - #30 #37 #39

### `PeerConnection` API:
* [x] Closing PeerConnection #40
* [x] Renegotiation - #55 #59
* [x] addTrack #38
* [x] removeTrack #49
* [x] Stopping transceivers #55
* [x] Tests for failing negotiation because of unsupported codecs - #48 #61

### Documentation:
* [x] Documentation and typespec improvements - #57
* [x] Transceiver guide - #56

# 0.2.0 (04.2024)

* [x] Initial benchmarks https://github.com/elixir-webrtc/bench/pull/1

### RTCP related:
* [x] RTCP Sender and Receiver Reports - #77 #78 #81
* [x] TWCC Feedbacks - #63
* [x] NACKs and RTX - #90 #101

### ICE features:
* [x] ICE mDNS (client-side) https://github.com/elixir-webrtc/ex_ice/pull/28 https://github.com/elixir-webrtc/ex_ice/pull/27
* [x] ICE relay candidates elixir-webrtc/ex_ice#33

# 0.3.0 (07.2024)

* [x] Inbound Simulcast #111
* [x] Benchmarks and profiling

## UX improvements/features:
* [x] User-friendly media/codec/extension configuration #120
* [x] ICE port range configuration https://github.com/elixir-webrtc/ex_ice/pull/43
* ~Better logs~

### [Fishjam](https://github.com/jellyfish-dev/jellyfish):
* ~Experimental WebRTC Endpoint using Elixir WebRTC~

### [Broadcaster](https://github.com/elixir-webrtc/apps/tree/master/broadcaster) improvements:
* [x] Viewer count https://github.com/elixir-webrtc/apps/pull/12
* ~Client-side ICE restarts~
* [x] Server-side stats https://github.com/elixir-webrtc/apps/pull/11
* [x] Simple chat https://github.com/elixir-webrtc/apps/pull/13

# 0.4.0 (08.2024)

### Website
* [x] A new website https://github.com/elixir-webrtc/elixir-webrtc.github.io/pull/2
[...]

PR #76, opened 2024-02-21
adding rtcp processing example

themusicman:
Wanted to contribute an example like the one in the Pion project here: https://github.com/pion/webrtc/tree/master/examples/rtcp-processing

### DIFF
--- a/examples/echo/lib/echo/peer_handler.ex
+++ b/examples/echo/lib/echo/peer_handler.ex
@@ -127,8 +127,6 @@ defmodule Echo.PeerHandler do
     {:ok, state}
   end
 
-  # RTCP is used for statistics and control information for media in WebRTC. 
-  # Using these packets you can get information about the quality of the media, round trip time and packet loss. 
   defp handle_webrtc_msg({:rtcp, _packets}, state) do
     # do something with RTCP packets
     {:ok, state}
--- a/examples/echo/lib/echo/peer_handler.ex
+++ b/examples/echo/lib/echo/peer_handler.ex
@@ -127,6 +127,13 @@ defmodule Echo.PeerHandler do
     {:ok, state}
   end
 
+  # RTCP is used for statistics and control information for media in WebRTC. 
+  # Using these packets you can get information about the quality of the media, round trip time and packet loss. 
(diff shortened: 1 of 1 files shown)


mickel8:
Hi @themusicman, thanks for the contribution!

A couple of notes:
* current example is almost the same as the `echo` example so we add a lot of code that we need to maintain and update as API changes
* handling RTCP in ex_webrtc is a little easier as we get those packets automatically. I can see why Pion has a separate example for this as they need to call ReadRTCP in a loop ([link](https://github.com/pion/webrtc/blob/master/examples/rtcp-processing/main.go#L39-L52))

I would favor adding a clause for RTCP in our `echo` example that just ignores those packets. We can add a short comment too. Something like this:

```elixir
  defp handle_webrtc_msg({:rtcp, _packets}, state) do
    # do something with RTCP packets
    {:ok, state}
  end
```

This way:
* we show how to receive RTCP packets
* we don't add too much extra code that needs to be maintained
* we don't send messages to the FE or log them in the console so the output is clean

themusicman:
@mickel8 Fair points. I updated the PR.

mickel8:
Thanks for the PR @themusicman! i have deleted the comment as we are going to talk about RTCP in our guides section (in hex docs) at some point. 

> Is there anything that you want help with in regards to the project? Maybe RTCP Sender and Receiver Reports?

That's something @LVala is already working on!
When it comes to the roadmap itself, I am working on ICE features (mDNS and TURN servers support) . 

Other things I have in my mind but are not planned for 0.2 (i.e. neither @LVala  nor me are going to work on them)
* support for AV1 codec, see #35 
* datachannels
* ex_webrtc_dashboard - more metrics, performance optimizations 
* jitter buffer 


and many more, I will update the list later this evening :wink:

themusicman:
@mickel8 I was thinking about suggesting datachannels too. I will look into that some and come up with a plan for implementation and present it to you. What do you think about that?

mickel8:
@themusicman Sounds great!

mickel8:
@themusicman I've created a new section in our Roadmap where I put things that can be addressed: https://github.com/elixir-webrtc/ex_webrtc/issues/28

PR #77, opened 2024-02-23
Add support for RTCP Receiver Reports

LVala:
This PR only adds `RTPReceiver.ReportRecorder` module, akin to `PeerConnection.TWCCRecorder`.
The actual functionality of sending the reports, as well as `RTPSender.ReportRecorder` for SenderReports, will be implemented in subsequent PRs.

### DIFF
--- a/test/ex_webrtc/rtp_receiver/report_recorder_test.exs
+++ b/test/ex_webrtc/rtp_receiver/report_recorder_test.exs
@@ -145,8 +145,8 @@ defmodule ExWebRTC.RTPReceiver.ReportRecorderTest do
         |> ReportRecorder.record_packet(packet, arrival_ts)
 
       # second packet arrived 1 millisecond late
-      # thus, jitter should be roughly equal to 1 millisecond in RTP ts units
-      assert_in_delta recorder.jitter, @clock_rate / 1000, 0.5
+      # thus, jitter should be roughly equal to 1 millisecond / 16 in RTP ts units
+      assert_in_delta recorder.jitter, @clock_rate / 1000 / 16, 0.5
 
       # remaining packets arrived perfectly on time
       # so the jitter should slowly converge to 0


lib/ex_webrtc/rtp_receiver/report_recorder.ex
```diff
@@ -0,0 +1,201 @@
...
+  @doc """
+  Creates an RTCP Receiver Report.
+  `time` parameter accepts output of `System.monotonic_time(:native)` as a value.
+  """
+  @spec get_report(t(), integer()) :: {ReceiverReport.t(), t()}
+  def get_report(recorder, time) do
+    received =
+      recorder.last_seq_no
+      |> seq_no_diff(recorder.last_report_seq_no)
+      |> min(@max_u24)
```
mickel8, line 84: Looking at definitions of `fraction_lost` and `cumulative number of packets lost`, we should calculate `received` from the beginning, not from the last sender report.  Is this done on purpose? I suspect that we follow libwebrtc or not really? 

Fraction lost is defined as 256 * fraction, where fraction is:

```
This fraction is defined to be the number
      of packets lost divided by the number of packets expected, as
      defined in the next paragraph.
```
      
and the number of packets expected:

```
The number of packets expected is
      defined to be the extended last sequence number received, as
      defined next, less the initial sequence number received.
```

lib/ex_webrtc/rtp_receiver/report_recorder.ex
```diff
@@ -0,0 +1,201 @@
...
+
+    lost =
+      recorder.lost_packets
+      |> MapSet.size()
+      |> min(@max_u24)
+
+    total_lost = min(recorder.total_lost + lost, @max_u24)
```
mickel8, line 91: We should count in duplicate packets here? 

```
This number is
      defined to be the number of packets expected less the number of
      packets actually received, where the number of packets received
      includes any which are late or duplicates.  Thus, packets that
      arrive late are not counted as lost, and the loss may be negative
      if there are duplicates.

```

test/ex_webrtc/rtp_receiver/report_recorder_test.exs
```diff
@@ -0,0 +1,212 @@
...
+      packet0 = %Packet{@packet | sequence_number: @seq_no - 10}
+      packet1 = %Packet{@packet | sequence_number: @seq_no - 6}
+      packet2 = %Packet{@packet | sequence_number: @seq_no - 3}
+      packet3 = %Packet{@packet | sequence_number: @seq_no - 1}
+
+      recorder =
+        @recorder
+        |> ReportRecorder.record_packet(packet1, @rand_ts)
+        # packet 0 will be ignored,
+        # see comment in rtp_receiver/report_recorder.ex:82
```
mickel8, line 85: Nitpick, referencing an exact line might outdate pretty quickly

mickel8 (commented):
Really nice code! 👍 Clean, concise and easy to understand. Also, very nice comment in the ReportRecorder 👀

I have some questions regarding fraction_lost and total_lost :thinking:

lib/ex_webrtc/rtp_receiver/report_recorder.ex
LVala: The very fist sentence of the definition in RFC 3550 is `The fraction of RTP data packets from source SSRC_n lost since the previous SR or RR packet was sent`. The `Appendix A.3` also explains how to calculate loss from the last reporting interval, not from the beginning. This 100% should be loss from the last reporting interval, also that's how every one implements it. 

When it comes to `cumulative number of packets lost`, we indeed use a number of packets lost since the very beginning and that's how it's implemented here.

lib/ex_webrtc/rtp_receiver/report_recorder.ex
LVala: Here I was just following what Pion does, if we wanted to include duplicate packets, then the whole `lost_packets` MapSet is unnecessary, we could just count the number of packets received. To be honest, I'm not really sure what is that, I'll just assumed that we shouldn't count duplicate packets, but I'll look into this more.

PR #83, opened 2024-03-05
Add local/remote ICE candidates to current/pending description.

LVala:
Thanks to this changes, session description obtained from `pending_local_description()` etc. will contain gathered/applied ICE candidates.

Adresses #80


lib/ex_webrtc/sdp_utils.ex
```diff
@@ -123,6 +123,13 @@ defmodule ExWebRTC.SDPUtils do
     |> Enum.map(fn {"candidate", attr} -> attr end)
   end
 
+  @spec add_ice_candidates(ExSDP.t(), [binary()]) :: ExSDP.t()
```
mickel8, line 126: To be consistent with get_ice_candidates?

```suggestion
  @spec add_ice_candidates(ExSDP.t(), [String.t()]) :: ExSDP.t()
```

test/ex_webrtc/peer_connection_test.exs
```diff
@@ -885,6 +885,27 @@ defmodule ExWebRTC.PeerConnectionTest do
     end)
   end
 
+  describe "get_description" do
```
mickel8, line 888: Nitpick: Could we put this after set_local_description test? I tried to match the order of tests with the order of functions in peer_connection.ex for easier navigation
```suggestion
  describe "get_local_description/1" do
```

mickel8 (approved):

PR #84, opened 2024-03-05
Allow for setting custom controlling process

mickel8:


test/ex_webrtc/peer_connection_test.exs
```diff
@@ -199,6 +199,19 @@ defmodule ExWebRTC.PeerConnectionTest do
 
   # API TESTS
 
+  test "controlling process" do
```
mickel8, line 202: Not sure if we want this test. I feel like it's pretty ugly :thinking:

LVala: I mean, if you have written it already...

LVala (approved):
SGTM, although do we have any particular need for this feature?

mickel8:
> SGTM, although do we have any particular need for this feature?

Yup, it's needed to allow for spawning PeerConnection under supervision tree

PR #86, opened 2024-03-11
Fix `TWCCRecorder` base timestamp evaluation bug

LVala:
The `TWCCRecorder` module has a default argument that keeps the initial value of `System.monotonic_time()` (so the subsequent calls could be offset to be positive, as output of `System.monotonic_time()` is usually negative). The value is evaluated at the compile time and differs from the runtime base `System.monotonic_time()` value, which is obviously undesired. This PR fixes this issue.


LVala:
@clone1018 most likely, I didn't recompile the module for some time, had to update Elixir version to notice the issue.

mickel8 (approved):
Works for me!

PR #87, opened 2024-03-11
Use normalized comparison in PeerConnection.Configuration.supported_codec?/2

clone1018:
This PR adds a couple of minor things:

* Add SDP offer from OBS to tests
* Lowercase mime type for codec comparison
* Show developer a warning whenever no codecs in the PeerConnection matched the mline


lib/ex_webrtc/rtp_transceiver.ex
```diff
@@ -3,6 +3,8 @@ defmodule ExWebRTC.RTPTransceiver do
   Implementation of the [RTCRtpTransceiver](https://www.w3.org/TR/webrtc/#dom-rtcrtptransceiver).
   """
 
+  require Logger
```
clone1018, line 6: Dunno how you all feel about Loggers inside libraries, let me know if you'd like a different approach but this warning is useful for implementing devs.

test/ex_webrtc/peer_sdp_test.exs
```diff
@@ -0,0 +1,41 @@
+defmodule ExWebRTC.BrowserSDPTest do
```
mickel8, line 1: Could we also update the README in our fixtures?

```suggestion
defmodule ExWebRTC.PeerSDPTest do
```

lib/ex_webrtc/rtp_transceiver.ex
mickel8: We put on warning things that are totally unexpected and the rest of things that are correct on debug so I would put the new log on debug :wink:

lib/ex_webrtc/rtp_transceiver.ex
```diff
@@ -89,6 +91,15 @@ defmodule ExWebRTC.RTPTransceiver do
   @spec from_mline(ExSDP.Media.t(), non_neg_integer(), Configuration.t()) :: t()
   def from_mline(mline, mline_idx, config) do
     codecs = get_codecs(mline, config)
+
+    if codecs == [] do
+      rtpmap = ExSDP.get_attribute(mline, :rtpmap)
+
+      Logger.warning(
+        "No valid codecs for \"#{rtpmap}\" found in registered %ExWebRTC.PeerConnection.Configuration{}"
```
mickel8, line 99: I think we should use inspect here, shouldn't we? rtpmap should be a struct.

```suggestion
        "No valid codecs for #{inspect(rtpmap)} found in registered %ExWebRTC.PeerConnection.Configuration{}"
```

mickel8 (changes requested):
Awesome, thanks! Just three small, minor fixes and we can merge!

lib/ex_webrtc/rtp_transceiver.ex
clone1018: This actually prints nicely by default thanks to: https://github.com/membraneframework/ex_sdp/blob/master/lib/ex_sdp/attribute/rtp_mapping.ex#L62

Let me know if you want to keep that, or swap it out for inspect!

PR #88, opened 2024-03-11
Add WHEP from file example

clone1018:
Adds an example of streaming from a file to multiple viewers via a WHEP negotiated WebRTC connection.


mickel8:
Thanks a lot!

PR #90, opened 2024-03-20
Generate NACKs and handle incoming RTX packets

LVala:


lib/ex_webrtc/rtp_receiver/nack_generator.ex
```diff
@@ -0,0 +1,110 @@
...
+
+  def record_packet(generator, packet) do
+    %__MODULE__{
+      lost_packets: lost_packets,
+      last_sn: last_sn,
+      max_nack: max_nack
+    } = generator
+
+    delta = packet.sequence_number - last_sn
+    in_order? = delta < -@breakpoint or (delta > 0 and delta < @breakpoint)
```
mickel8, line 46: Maybe we could move this to a separate RTP utils module

mickel8 (approved):

lib/ex_webrtc/rtp_receiver/nack_generator.ex
LVala: I wouldn't, the amount of repetition is nearly non-existent and moving a single-line function to a separate module would make it harder to read, arguably.

PR #92, opened 2024-03-21
Allow `add_ice_candidate` when `current_remote_description` is not set

LVala:
This PR fixes a bug that makes it impossible to use `add_ice_candidate` with only pending remote description.

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -433,9 +433,13 @@ defmodule ExWebRTC.PeerConnection do
   end
 
   @impl true
-  def handle_call({:add_ice_candidate, _}, _from, %{current_remote_desc: nil} = state) do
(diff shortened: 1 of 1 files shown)


lib/ex_webrtc/peer_connection.ex
```diff
@@ -433,9 +433,13 @@ defmodule ExWebRTC.PeerConnection do
   end
 
   @impl true
-  def handle_call({:add_ice_candidate, _}, _from, %{current_remote_desc: nil} = state) do
-    {:reply, {:error, :no_remote_description}, state}
-  end
+  def handle_call(
+        {:add_ice_candidate, _},
+        _from,
+        %{current_remote_desc: crd, pending_remote_desc: prd} = state
```
mickel8, line 439:
```suggestion
        %{current_remote_desc: nil, pending_remote_desc: nil} = state
```

mickel8 (approved):

PR #102, opened 2024-04-20
Adjustments for BUNDLE group. Update ex_ice.

mickel8:
Besides fixes listed below, this PR also update deps and adjusts code to the newest ex_ice.

Fixes: 

* an m-line is only rejected when it has port set to 0 and there is no `bundle-only` attribute, see [here](https://www.rfc-editor.org/rfc/rfc8843#name-sdp-bundle-only-attribute). This is used by Firefox when PC is configured with `bundlePolicy: "maxBundle"` 
* rejected m-lines are not included in the BUNDLE group, see [here](https://www.rfc-editor.org/rfc/rfc8829#section-5.2.2-10) 
* in `maxBundle` policy (the only one we support), candidates are only included in the first m-line, see [here](https://www.rfc-editor.org/rfc/rfc8829#section-4.1.1-8.6). This seems to be true even if the first m-line is stopped in subsequent offers, at least until ICE restart, which I left untouched i.e. we always include candidates in the first mline 

Questionable:

* in `maxBundle` policy, all mlines but the first one should have `bundle-only` attribute added, and port in the m-section set to 0, see [here](https://www.rfc-editor.org/rfc/rfc8829#section-4.1.1-8.6). However, there are differences between JSEP RFC, BUNDLE RFC, and browser implementations (see [RFC 8829. sec. 1.3](https://datatracker.ietf.org/doc/html/rfc8829#section-1.3) and because of that, we still don't use neither port 0 nor bundle-only attribute 

* depending on the `bundle-policy`, browsers may send candidates for every m-section, only for the first m-section, or for every type of m-section. There is a question, which candidates we should add to the ICE agent on the answerer side (as we only have one ICE stream and one ICE component) - all of them or only those from the first m-section. Looking at [RFC 8829, sec. 5.11](https://datatracker.ietf.org/doc/html/rfc8829#section-5.11):

  > If the answer contains valid bundle groups, discard any ICE components for 
  > the "m=" sections that will be bundled onto the primary ICE components in 
  > each bundle, and begin muxing these "m=" sections accordingly, as described
  > in [RFC8843], Section 7.4.

  so we could take only those candidates that belong to the first m-section as other ICE 
  components on the offerer side should be closed anyway. However, because current solution seems to work, I would 
  go back to this question once we notice any problems.

* depending on the `rtcpMuxPolicy` browsers may generate candidates only for one ICE component (RTP) or for two ICE components (RTP and RTCP) doubling the number of ICE candidates. There is the same question as in the point above, and we do exactly the same thing i.e. we add all candidates regardless of their ICE components.


lib/ex_webrtc/sdp_utils.ex
```diff
@@ -78,6 +66,26 @@ defmodule ExWebRTC.SDPUtils do
     end
   end
 
+  @spec get_media_direction(ExSDP.Media.t()) ::
+          :sendrecv | :sendonly | :recvonly | :inactive | nil
+  def get_media_direction(media) do
```
mickel8, line 71: Moved to be close to other `get_*` functions

test/ex_webrtc/peer_connection_test.exs
```diff
@@ -247,12 +247,12 @@ defmodule ExWebRTC.PeerConnectionTest do
...
 
-      "a=" <> desc_cand =
+      desc_cands =
         desc.sdp
         |> String.split("\r\n")
-        |> Enum.find(&String.starts_with?(&1, "a=candidate:"))
+        |> Enum.filter(&String.starts_with?(&1, "a=candidate:"))
 
-      assert desc_cand == cand.candidate
+      assert ("a=" <> cand.candidate) in desc_cands
```
mickel8, line 255: We no longer can relay on candidates order as they are stored as map in ex_ice

PR #111, opened 2024-05-29
Inbound Simulcast

LVala:


lib/ex_webrtc/rtp_receiver/simulcast_demuxer.ex
```diff
@@ -0,0 +1,94 @@
...
+  alias ExRTP.Packet.Extension.SourceDescription
+
+  @rid_uri "urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id"
+  @rrid_uri "urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id"
+
+  @type t() :: %__MODULE__{
+          rid_ext_id: non_neg_integer() | nil,
+          rrid_ext_id: non_neg_integer() | nil,
+          ssrc_to_rid: %{(ssrc :: non_neg_integer()) => rid :: String.t()}
+        }
```
mickel8, line 17: Our demuxer does not check whether received rid was specified in SDP offer so in theory, if in SDP we have `h,m,l` and we receive a packet with e.g. `a` we will handle such packet. Am I correct?

lib/ex_webrtc/peer_connection/configuration.ex
```diff
@@ -60,10 +68,13 @@ defmodule ExWebRTC.PeerConnection.Configuration do
                                   {extmap.uri, extmap}
                                 end)
 
-  @mandatory_video_rtp_hdr_exts Map.new([:mid, :twcc], fn ext_shortcut ->
-                                  extmap = Map.fetch!(@rtp_hdr_extensions, ext_shortcut).ext
-                                  {extmap.uri, extmap}
-                                end)
+  @mandatory_video_rtp_hdr_exts Map.new(
+                                  [:mid, :twcc, :rtp_stream_id, :repaired_rtp_stream_id],
```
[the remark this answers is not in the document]
LVala: It is ok, although not necessary. Otherwise, the user will need to add the extension themselves, which is very error-prone and hard to debug. I would leave it as it is at least until we tackle the codec/extension config in general.

lib/ex_webrtc/rtp_receiver/simulcast_demuxer.ex
LVala: Yeah, I decided it isn't much of an issue.

mickel8 (approved):

PR #114, opened 2024-06-03
Fix send_rtp for non-existing track id

ndrean:
I did not returned a `{:stop, :unrecognized_id, state}` as I don't know how you want to treat this.

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -997,6 +997,8 @@ defmodule ExWebRTC.PeerConnection do
           "Attempted to send packet to track with unrecognized id: #{inspect(track_id)}"
         )
 
+        {:stop, :invalid_track_id, state}
+
       {transceiver, idx} ->
         {packet, state} =
           case Map.fetch(state.config.video_rtp_hdr_exts, @twcc_uri) do
@@ -1022,8 +1024,6 @@ defmodule ExWebRTC.PeerConnection do
 
         transceivers = List.replace_at(state.transceivers, idx, transceiver)
         state = %{state | transceivers: transceivers}
-
-        {:noreply, state}
 
         {:noreply, state}
     end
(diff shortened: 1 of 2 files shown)


mickel8 (commented):
Hi @ndrean,

thanks for the PR!

Could you please add a test for this too?

ndrean:
Yes....indeed 


<https://github.com/elixir-webrtc/ex_webrtc/blob/cbfdc56b5d3b8f2e62cfbc832786fdd653827a16/test/ex_webrtc/peer_connection_test.exs#L953>

`test_send_data(pc1, pc2, track1, track1)` will make it fail, but... I will be verbose.

Does the test written above makes sense as such? No, but, for which reason else than bad code would it fail? I did not figure out why you made this case.

We should  `{:stop, :invalid_track_id, state}` but I don't know how to capture the failure in the test. The `PC.start_link` propagates it to the main test process but I can't figure out what the main process is, so can't supervise it.

So I didn't added a PR (which should also  remove the second `{:noreply, state}`).

The context. I want to pass data between 2 clients A and B connected to a ExWebRTC server, that is A<->pcA, B<->pcB (WebSockets, not HTTP). I just tried quickly,  in the callback `handle_info(pcA, {:rtp, tr_id, packet})` where I expect to receive A's tracks,  to do a  `PC.send_rtp(get_pcB, tr_id, packet)`. This fails for the (future)  "non-existing" reason.
Whatever, this is fundamentally wrong, perhaps because there is no RTC pcA<->pcB so no transitivity?  but is another subject.

ndrean:
`:EXIT` trapped in fact with `Process.flag(:trap_exit, true)`. Discovered this. Seems to pass.

mickel8:
@ndrean thanks for your work!

I have modified a little bit the behavior so that pc doesn't stop when we try to send data on invalid track id. 

I also created a separate test case for send_rtp as I belive it fits better :slightly_smiling_face:

ndrean:
Thanks both of you for your work!

PR #120, opened 2024-06-20
Improve UX of `Configuration`

LVala:
With this PR:
- Users can configure whether they want to use features like Sender/Receiver Reports, TWCC, RTX, etc.
-  adds necessary RTP header extensions, RTCP feedbacks, and RTX codecs by default depending on features enabled
- in general, created SDPs correspond to the config more closely


lib/ex_webrtc/peer_connection.ex
```diff
@@ -717,67 +722,58 @@ defmodule ExWebRTC.PeerConnection do
...
     end
   end
 
   @impl true
   def handle_call({:stop_transceiver, tr_id}, _from, state) do
-    idx = Enum.find_index(state.transceivers, fn tr -> tr.id == tr_id end)
+    state.transceivers
+    |> Enum.with_index()
+    |> Enum.find(fn {tr, _idx} -> tr.id == tr_id end)
+    |> case do
```
mickel8, line 746: While those changes are okay I feel like this is only refactoring? If yes, in general, I would put such changes in a separate PR :wink:

mickel8 (commented):
Nice, really good job! I just put a couple of nitpicks

lib/ex_webrtc/peer_connection/configuration.ex
```diff
@@ -22,11 +29,6 @@ defmodule ExWebRTC.PeerConnection.Configuration do
       mime_type: "video/VP8",
       clock_rate: 90_000
     },
-    %RTPCodecParameters{
-      payload_type: 45,
-      mime_type: "video/AV1",
-      clock_rate: 90_000
-    },
```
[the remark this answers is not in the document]
LVala: By mistake

lib/ex_webrtc/peer_connection.ex
LVala: sure, I came to the same conclusion, but it's not much so I left it here

lib/ex_webrtc/peer_connection/configuration.ex
```diff
@@ -124,9 +129,14 @@ defmodule ExWebRTC.PeerConnection.Configuration do
   * `audio_codecs` and `video_codecs` - lists of audio and video codecs to negotiate. By default these are equal to
   `default_audio_codecs/0` and `default_video_codecs/0`. To extend the list with your own codecs, do
   `audio_codecs: Configuration.default_audio_codecs() ++ my_codecs`.
-  * `header_extensions` - list of RTP header extensions to negotiate. Refer to `t:header_extension/0` for more information.
-  * `feedbacks` - list of RTCP feedbacks to negotiate. Refer to `t:rtcp_feedback/0` for more information.
   * `features` - feature flags for some of the ExWebRTC functinalities. Refer to `t:feature/0` for more information.
+  * `rtp_header_extensions` - list of RTP header extensions to negotiate. Refer to `t:rtp_header_extension/0` for more information.
+  * `rtcp_feedbacks` - list of RTCP feedbacks to negotiate. Refer to `t:rtcp_feedback/0` for more information.
+
+  Instead of manually enabling an RTP header extension or an RTCP feedback, you may want to use a `t:feature/0` instead, which will enable
```
mickel8, line 136:
```suggestion
  Instead of manually enabling an RTP header extension or an RTCP feedback, you may want to use a `t:feature/0`, which will enable
```

mickel8 (approved):

PR #121, opened 2024-06-21
Fix simulcast issues

LVala:
This PR:
- fixes `send_pli` not working with simulcast
- fixes duplicate id of `inbound_rtp` sections in `get_stats`

### DIFF
--- a/lib/ex_webrtc/rtp_receiver.ex
+++ b/lib/ex_webrtc/rtp_receiver.ex
@@ -65,7 +65,7 @@ defmodule ExWebRTC.RTPReceiver do
       track: track,
       codec: codec,
       simulcast_demuxer: SimulcastDemuxer.new(rtp_hdr_exts),
-      layers: %{nil => init_layer(codec)}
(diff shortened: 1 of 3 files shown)


lib/ex_webrtc/rtp_receiver.ex
```diff
@@ -213,8 +213,10 @@ defmodule ExWebRTC.RTPReceiver do
   @spec get_stats(receiver(), non_neg_integer()) :: [map()]
   def get_stats(receiver, timestamp) do
     Enum.map(receiver.layers, fn {rid, layer} ->
+      id = if(rid == nil, do: receiver.track.id, else: "#{receiver.track.id}, rid: #{rid}")
```
mickel8, line 216: I would favor using `:` and no spaces as puting spaces and commas in id isn't a common practice

```suggestion
      id = if(rid == nil, do: receiver.track.id, else: "#{receiver.track.id}:#{rid}")
```

mickel8 (approved):

PR #126, opened 2024-07-01
Add `RTP.Munger` and H264 codec utilities

LVala:


lib/ex_webrtc/rtp/h264.ex
```diff
@@ -0,0 +1,53 @@
+defmodule ExWebRTC.RTP.H264 do
+  @moduledoc """
+  Utilities for RTP packets carrying H264 encoded payload.
+  """
+
+  alias ExRTP.Packet
+
+  # Copied nearly 1-to-1 from https://github.com/membraneframework/membrane_rtp_h264_plugin/blob/master/lib/rtp_h264/utils.ex
```
mickel8, line 8: Can we please copy relevant comment from Membrane RTP plugin to make this chain a litte shorter? Just mention that we took the code from galene and it is unclear why we check against nalu type 5 and 7 instead of just one of them

lib/ex_webrtc/rtp/munger.ex
```diff
@@ -0,0 +1,177 @@
...
+      {:noreply, %{state | munger: munger}}
+    else
+      {:noreply, state}
+    end
+  end
+
+  def handle_info({:change_layer, layer}, state) do
+    munger = Munger.update(state.munger)
+    {:noreply, %{munger: munger, layer: layer}
+  end
```
mickel8, line 31: I belive that in most cases, `rid` is denoted using a single letter so I would keep the convention

```suggestion
  # assume you receive two layers:  "h" ("high") and "l" ("low")
  # and this is a GenServer

  def init() do
    {:ok, %{munger: Munger.new(90_000), layer: "h"}}
  end

  def handle_info({:ex_webrtc, _from, {:rtp, _id, rid, packet}}, state) do
    if rid == state.layer do
      {packet, munger} = Munger.munge(state.munger, packet)
      send_packet_somewhere(packet)
      {:noreply, %{state | munger: munger}}
    else
      {:noreply, state}
    end
  end

  def handle_info({:change_layer, layer}, state) do
    # indicate to the munger that the next packet will be from a new layer
    munger = Munger.update(state.munger)
    {:noreply, %{munger: munger, layer: layer}
  end
```

test/ex_webrtc/rtp/munger_test.exs
```diff
@@ -0,0 +1,166 @@
...
+    {packet, munger} = Munger.munge(munger, l2_packet)
+    assert packet.timestamp in 5001..6000
+    ts = packet.timestamp
+
+    l2_packet = %{@packet | sequence_number: 201, timestamp: 1000}
+    {packet, _munger} = Munger.munge(munger, l2_packet)
+    assert packet.timestamp == ts + 1001
+  end
+
+  test "handle output timestamp rollover" do
```
mickel8, line 154:
```suggestion
  test "handles output timestamp rollover" do
```

mickel8 (approved):
Really good job! :rocket:

PR #140, opened 2024-07-18
Add `track_id` to packets sent in `{:rtcp, ...}` message

### LINKED ISSUE #138: Unable to demux incoming RTCP packets
Currently (after the changes to API which have hidden most of the transceiver's internal state) it is impossible to demux incoming RTCP packets (PLI specifically), as the only way to learn about tracking SSRC is to track incoming packets, which is very inconvenient.

We should demux RTCP packets similarly to what we do with RTP packets. Some packets (like TWCC) are "undemuxable", these can be just forwarded as it is done now.

LVala:
Close #138

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -65,8 +65,9 @@ defmodule ExWebRTC.PeerConnection do
   * `:track_muted`, `:track_ended` - these match the [MediaStreamTrack events](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack#events).
   * `:rtp` and `:rtcp` - these contain packets received by the PeerConnection. The third element of `:rtp` tuple is a simulcast RID and is set to `nil` if simulcast
   is not used.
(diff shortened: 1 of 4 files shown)


lib/ex_webrtc/peer_connection.ex
```diff
@@ -78,7 +81,7 @@ defmodule ExWebRTC.PeerConnection do
            | {:track_muted, MediaStreamTrack.id()}
            | {:track_ended, MediaStreamTrack.id()}
            | {:rtp, MediaStreamTrack.id(), String.t() | nil, ExRTP.Packet.t()}}
-          | {:rtcp, [ExRTCP.Packet.packet()]}
+          | {:rtcp, [{ExRTCP.Packet.packet(), MediaStreamTrack.id() | nil}]}
```
mickel8, line 84:
```suggestion
          | {:rtcp, [{MediaStreamTrack.id() | nil, ExRTCP.Packet.packet()}]}
```

mickel8 (approved):

PR #141, opened 2024-07-23
Further improvements to `Introduction to Elixir WebRTC` tutorial

LVala:


guides/introduction/negotiation.md
```diff
@@ -2,31 +2,29 @@
 
 Before starting to send or receive media, you need to negotiate the WebRTC connection first, which comes down to:
```
mickel8, line 3: Really nice! Clean and conciese :+1: 

The only thing is that I feel like we have too many admonitions in the `Offer and answer exchange section`. They distract  a  reader, don't they?

I am wondering about adding at its end something like this:

```
Notes:
* here admonition about `PeerConnection can be bidirectional`
* here admonition about `Renegotiations`
```

I think I would also inline `Offers, answers and SDP` admoniton and refactor it just a little to mention that answer  can reject incoming offer or its parts?

```
Finally, we have to create and set an offer.
It will contain information on how many tracks we want to send, which codecs we want to use, whether we are willing to also receive something or not and so on.
The other side responds with an answer, which can either accept, reject or partially accept our offer (e.g. accept only audio tracks).
Both offer and answer are carried in a text format called SDP. You can read more about it in the SDP Anatomy tutorial from webrtcHacks.
```

Any thoughts?

guides/introduction/negotiation.md
```diff
@@ -115,75 +115,99 @@ Now we create the answer, set it, and send it back to the web browser.
...
+> Remember when we created the `RTCPeerConnection` object at the beginning of this tutorial? It was configured
+with `iceServers` options:
+>
+> ```js
+> const pc = new RTCPeerConnection({ iceServers: "stun:stun.l.google.com:19302" })
+> ```
+>
+> It is a list of STUN/TURN servers that the PeerConnection will try to use. You can learn more about
+> it in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection) but
+> it boils down to the fact that lack of any STUN servers might cause you trouble connecting with other peers, so make sure
```
mickel8, line 209: I think that it will be easier to mention that iceServers are used to discover more ICE candidates when we are behind NAT.

You think about it as: okay ICE generates some candidates but those might not be enough so we provide external servers to discover more of them

guides/introduction/forwarding.md
```diff
@@ -39,45 +39,36 @@ flowchart LR
...
-    # out_tracks (tracks we will send media to) = %{kind => id}
-    out_tracks = %{audio: audio_track.id, video: video_track.id}
-    {:ok, %{pc: pc, out_tracks: out_tracks, in_tracks: %{}}}
-  end
-
-  # ...
+def init(_) do
+  {:ok, pc} = PeerConnection.start_link(ice_servers: [%{urls: "stun:stun.l.google.com:19302"}])
+
+  # we expect to receive two tracks from the web browser - one for audio, one for video
```
mickel8, line 49: Nitpick, to me this sounds better 

```suggestion
  # we expect to receive two tracks from the web browser - one audio and one video
```

guides/introduction/forwarding.md
```diff
@@ -39,45 +39,36 @@ flowchart LR
...
+  # so we also need to add two tracks here, we will use these to forward media
+  # from each of the web browser tracks
```
mickel8, line 51: Nitpick, I would write:

```suggestion
  # so we also need to add two tracks here, we will use them to loop media back to the browser
```

guides/introduction/consuming.md
```diff
@@ -1,27 +1,123 @@
 # Consuming media data
 
-Other than just forwarding, we probably would like to be able to use the media right in the Elixir app to
-e..g feed it to a machine learning model or create a recording of a meeting.
+Other than just forwarding, we would like to be able to use the media right in the Elixir app to e.g.
+use it as a machine learning model input, or create a recording of a meeting.
 
-In this tutorial, we are going to build on top of the simple app from the previous tutorial by, instead of just sending the packets back, depayloading and decoding
-the media, using a machine learning model to somehow augment the video, encode and payload it back into RTP packets and only then send it to the web browser.
+In this tutorial, we are going to depayload and decode received video data to use it for ML inference.
```
mickel8, line 6: We have already introduced RTP term in the previous tutorial so maybe let's not introduce another term just at the beginning?


```suggestion
In this tutorial, we are going to unpack RTP to get the actual audio and video frames ready to be decoded and used for ML inference.
```

guides/introduction/consuming.md
```diff
@@ -1,27 +1,123 @@
...
-> In WebRTC, most likely you will encounter VP8, H264 or AV1 video codecs and Opus audio codec. Codecs that will be used during the session are negotiated in
-> the SDP offer/answer exchange. You can tell what codec is carried in an RTP packet by inspecting its payload type (`packet.payload_type`,
-> a non-negative integer field) and match it with one of the codecs listed in this track's transceiver's `codecs` field (you have to find
-> the `transceiver` by iterating over `PeerConnection.get_transceivers` as shown previously in this tutorial series).
+> In WebRTC, most likely you will encounter VP8, H264 or AV1 video codecs and Opus audio codec. Codecs used during the session are negotiated in
+> the SDP offer/answer exchange. You can tell what codec is carried in an RTP packet by inspecting its payload type (`payload_type` field in the case of Elixir WebRTC).
+> This value should correspond to one of the codecs included in the SDP offer/answer.
 
-_TBD_
+Unfortunately, in other cases, we need to do more work. In video, things are more complex: each video frame is usually split into multiple packets (and
```
mickel8, line 22:
```suggestion
In video, things are more complex: each video frame is usually split into multiple packets (and
```

guides/introduction/consuming.md
```diff
@@ -1,27 +1,123 @@
...
 
-## Deplayloading RTP
+## Depayloading RTP
 
-We refer to the process of taking the media payload out of RTP packets as _depayloading_.
+We refer to the process of getting the media payload out of RTP packets as _depayloading_. It may seem straightforward at first,
+we just take the payload of the packets and we get a stream of media data. Sometimes it is that simple, like in the
+case of Opus-encoded audio, where each of the RTP packets is, more or less, 20 milliseconds of audio, and that's it.
 
 > #### Codecs {: .info}
```
mickel8, line 14: What about inlining this and changing the order of the whole section a little bit?
Maybe let's start with explaining what browser does i.e.

```
When browser sends audio or video it does the following steps:
* aquire audio or video frame
* encode it so it takes less space and ueses less bandwidth 
* depending on its size, pack into a single or multiple RTP packets 

When we receive audio or video from a web browser, we have to do the opposite:
* collect multiple RTP packets that contain a single audio or a single video frame
* unpack this frame from RTP packets
* decode it

In practice, audio frames always fit into a single RTP packet.
In case of video, this depends on its resolution and compression level.
However, you don't have to worry about this. 
Just feed RTP packets into so called depayloader and it will return you a frame once it collect all needed data.
```

guides/introduction/negotiation.md
LVala: I inlined the `Offers, answers and SDP` and `Renegotations`.

mickel8 (approved):

PR #147, opened 2024-08-06
Add behaviour and dynamic dispatch for (de)payloaders

### LINKED ISSUE #143: Create a common behaviour for payloaders and depayloaders

sgfn:
Closes #143

### DIFF
--- a/examples/save_to_file/lib/save_to_file/peer_handler.ex
+++ b/examples/save_to_file/lib/save_to_file/peer_handler.ex
@@ -10,7 +10,7 @@ defmodule SaveToFile.PeerHandler do
   }
 
   alias ExWebRTC.Media.{IVF, Ogg}
-  alias ExWebRTC.RTP.{Opus, VP8}
+  alias ExWebRTC.RTP.Depayloader
 
   @behaviour WebSock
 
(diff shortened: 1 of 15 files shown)


LVala (commented):
I'm a bit unsure about adding specs to callback implementation. I believe that `ex_doc` will generate the spec anyway (based on the callback) and making the callback and implementation spec different is a big yucky, but on the other hand, we know that e.g. `Opus.depayload` will never return `{nil, depyayloader}`. I'm a bit torn when it comes to this.

LVala:
Removing the `@doc` will automatically add a doc with "Callback implementation for <link to docs for the callback>". Indeed, not `@spec` in such a case.

Not sure which one is better, but having explicit spec seems more useful.

mickel8:
I think that at the end of the day, we don't want to document functions from Opus/VP8 payloaders/depaylaoders. Instead, the whole documentation should be in the behaviour module or in the module that is responsible for dispatching or in both. 

E.g.

```elixir
d = ExWebRTC.RTP.Depayloader.new(ExWebRTC.RTP.Depayloader.VP8, some_options)
{frame, d} = ExWebRTC.RTP.Depayloader.depayload(d, packet)
```

In such case, we only need to document ExWebRTC.RTP.Depayloader.VP8 module.  Its functions shouldn't appear in docs. But this is the final outcome.

mickel8 (approved):

sgfn:
@mickel8 Could you please reiterate on how exactly do you suggest the docs should look like? Specifically, which parts should I get rid of?

mickel8:
@sgfn please take a look at #154

PR #157, opened 2024-08-20
Add DataChannels

LVala:
Rough, but working first version of DataChannel support. I'll take care of `TODO`s, applying negotiated channel parameters and closing the channels in subsequent PRs.


sgfn (commented):
LGTM (I got lost about 1/3 through)

test/ex_webrtc/data_channel_test.exs
```diff
@@ -0,0 +1,170 @@
...
+
+      :ok = negotiate(pc1, pc2)
+      :ok = connect(pc1, pc2)
+
+      label = "my label"
+      {:ok, %DataChannel{ref: ref}} = PeerConnection.create_data_channel(pc2, label)
+
+      # NOTICE: in theory, this should be true, but as of now, we perform the SCTP handshake
+      # even if the SDP did not contain any data lines
+      # refute_receive {:ex_webrtc, ^pc1, {:data_channel, _}}
```
mickel8, line 102: Does it mean that we can setup datachannel without its prior negotiation? If yes, I assume this is not in the spec?

test/ex_webrtc/data_channel_test.exs
```diff
@@ -0,0 +1,170 @@
...
+      :ok = PeerConnection.send_data(pc1, ref1, data1)
+      assert_receive {:ex_webrtc, ^pc2, {:data, ^ref2, ^data1}}
+
+      data2 = for i <- 1..1024, into: <<>>, do: <<i>>
+      :ok = PeerConnection.send_data(pc1, ref1, data2)
+      assert_receive {:ex_webrtc, ^pc2, {:data, ^ref2, ^data2}}
+
+      :ok = PeerConnection.send_data(pc1, ref1, <<>>)
+      assert_receive {:ex_webrtc, ^pc2, {:data, ^ref2, <<>>}}
+    end
```
mickel8, line 137: Is there any reason why we send so many messages? If yes, then it would be nice to add a comment

mickel8 (commented):
Can't wait when we announce this feature! :rocket:

lib/ex_webrtc/data_channel.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  @typedoc """
+  Struct representing the DataChannel.
+
+  All of the fields have the same meaning as in [RTCDataChannel](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel)
+  except for `ref` which is a local identifier used when refering to this DataChannel in
+  received messages or when calling `ExWebRTC.PeerConnection.send_data/3` function.
+  """
+  @type t() :: %__MODULE__{
+          ref: ref(),
```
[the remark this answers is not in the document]
LVala: This is very much not internal, we cannot use id as DataChannels are assigned dynamically when doing the "DataChannel handshake". This means that DataChannels created before SCTP has been connected, `id` is always nil. Also, ids can be reused so we need to make a distinction between new and old channels with the same id.

I initially did use ids, but later found out about these and needed to use something else, refs seem nice.

lib/ex_webrtc/sctp_transport/dcep.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  defmodule DataChannelAck do
+    @moduledoc false
+
+    defstruct []
+
+    def decode(<<>>), do: {:ok, %__MODULE__{}}
+    def decode(_other), do: :error
+
+    def encode(%__MODULE__{}), do: <<0x02>>
```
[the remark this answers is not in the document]
LVala: The first byte indicates the PPI (type of the packet) and we match it in DCEP.decode.

The specific packet modules' `decode` function only gets the rest, and in the case of `Ack`, the rest is empty. When encoding, we include the identifier PPI byte.

test/ex_webrtc/data_channel_test.exs
LVala: I wouldn't consider 3 to be _so many_. Each tests a packet with different properties, which is nice considering that empty packets are handled differently. I should've made the big packet even bigger to get over SCTP message size limit.

lib/ex_webrtc/peer_connection.ex
```diff
@@ -1293,13 +1401,17 @@ defmodule ExWebRTC.PeerConnection do
...
-        case Enum.find(transceivers, &(&1.mline_idx == idx)) do
-          # if there is no transceiver, the mline must have been rejected
-          # in the past (in the offer or answer) so we always set the port to 0
-          nil ->
+        tr = Enum.find(transceivers, &(&1.mline_idx == idx))
+
+        cond do
+          SDPUtils.data_channel?(local_mline) ->
+            {:mid, mid} = ExSDP.get_attribute(local_mline, :mid)
+            generate_data_mline(mid, opts)
```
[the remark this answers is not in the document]
LVala: Yeah, I guess so (it needs different ICE creads, fingerprint, potentially `sctp-port`, DTLS direction, so we cannot use the same one as in the remote description.

examples/chat/priv/static/script.js
```diff
@@ -0,0 +1,53 @@
+const pcConfig = { 'iceServers': [{ 'urls': 'stun:stun.l.google.com:19302' },] };
+const chatInput = document.getElementById("chatInput");
+const chatMessages = document.getElementById("chatMessages");
+
+const proto = window.location.protocol === "https:" ? "wss:" : "ws:"
+const ws = new WebSocket(`${proto}//${window.location.host}/ws`);
+ws.onopen = _ => start_connection(ws);
+ws.onclose = event => console.log("WebSocket connection was terminated:", event);
```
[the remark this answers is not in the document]
LVala: Hmm, then I should change this in all of the examples

test/ex_webrtc/data_channel_test.exs
LVala: fixed

test/ex_webrtc/data_channel_test.exs
mickel8: Sorry, I misread it, thought we send 1024 :man_facepalming:

mickel8 (approved):

PR #158, opened 2024-08-20
Add jitter buffer

sgfn:
The battle-tested (I would assume) implementation from Membrane RTP Plugin


lib/ex_webrtc/rtp/jitter_buffer.ex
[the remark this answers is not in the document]
sgfn: Changed. Please have a look at the new API

lib/ex_webrtc/rtp/jitter_buffer.ex
```diff
@@ -0,0 +1,166 @@
...
+  @typedoc """
+  The 3-element tuple returned by all functions other than `new/1`.
+
+  * `buffer` - `t:#{inspect(__MODULE__)}.t/0`.
+  * `packets` - a list with packets flushed from the buffer as a result of the function call. May be empty.
+  * `timer_duration_ms` - see `t:timer/0`.
+
+  Generally speaking, all results of this type can be handled in the same way.
+  """
+  @type result :: {buffer :: t(), packets :: [Packet.t()], timer_duration_ms :: timer()}
```
LVala, line 37: Generally, the structure you're modifying is returned as the last thing (like in `Map.pop` for instance. We should probably stick to this convention

lib/ex_webrtc/rtp/jitter_buffer.ex
```diff
@@ -0,0 +1,166 @@
...
+
+  @doc """
+  Handles the end of a previously set timer.
+  """
+  @spec handle_timer(t()) :: result()
+  def handle_timer(buffer) do
+    %__MODULE__{buffer | state: :timer_not_set} |> send_packets()
+  end
+
+  @spec try_insert_packet(t(), Packet.t()) :: {:ok | :error, t()}
```
LVala, line 108: `spec`s for private functions? fancy

LVala (approved):
Have @mickel8 take a look at the API and we are probably good to go

lib/ex_webrtc/rtp/jitter_buffer.ex
sgfn: Might as well

lib/ex_webrtc/rtp/jitter_buffer.ex
```diff
@@ -0,0 +1,166 @@
...
+      |> records_to_packets()
+
+    {packets, nil, %__MODULE__{latency: buffer.latency}}
+  end
+
+  @doc """
+  Handles the end of a previously set timer.
+  """
+  @spec handle_timer(t()) :: result()
+  def handle_timer(buffer) do
```
mickel8, line 104: Nitpick, I believe that in most places we use `handle_timeout`
```suggestion

  def handle_timeout(buffer) do
```

test/ex_webrtc/rtp/jitter_buffer_test.exs
```diff
@@ -0,0 +1,127 @@
...
+        buffer.store
+        | flush_index: flush_index,
+          highest_incoming_index: flush_index
+      }
+
+      {:ok, store} = PacketStore.insert_packet(store, packet)
+      buffer = %{buffer | store: store}
+      {[^packet], nil, buffer} = JitterBuffer.flush(buffer)
+
+      assert buffer.store == %PacketStore{}
```
mickel8, line 124: You can call JitterBuffer.flush(buffer) for the second time and assert that packets is an empty list

test/ex_webrtc/rtp/jitter_buffer_test.exs
```diff
@@ -0,0 +1,127 @@
+defmodule ExWebRTC.RTP.JitterBufferTest do
```
mickel8, line 1: I feel like we inspect packet store in those tests too many times. 

E.g. instead of using PacketStore.dump use JitterBuffer.flush, instead of PacketStroe.insert_packet use JitterBuffer.place_packet

mickel8 (changes requested):

lib/ex_webrtc/rtp/jitter_buffer.ex
```diff
@@ -0,0 +1,166 @@
+defmodule ExWebRTC.RTP.JitterBuffer do
```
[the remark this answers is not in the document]
LVala: In the case of saving to a file, it might not make a difference. Still, when a player is on the receiving end, assuming we returned the very first packets right away, the player would need to handle the possibility of a `latency` long gap between two packets (so `latency` long freeze). When the initial latency is forced, we might store a bunch of packets that came earlier than the initial latency and then dump them all at once after the initial latency, but as long as there are no gaps longer than `latency`, no freezes will occur and RTP timestamps pace the playout anyway.

Does it make sense?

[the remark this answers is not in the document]
mickel8: Hmmmmm. What about this: if we don't postpone the playback at the beginning, we will do this on the first detected gap. So we either wait at the beginning or during the first detected gap. Is that the whole difference?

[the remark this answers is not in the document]
LVala: Ye, but you'd rather have an initial latency than freezes through the playback. Lack of the initial latency kinda defeats the whole purpose of a jitter buffer.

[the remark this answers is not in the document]
mickel8: Agree 👍

test/ex_webrtc/rtp/jitter_buffer/packet_store_test.exs
```diff
@@ -0,0 +1,295 @@
...
+      assert has_packet(store, packet)
+    end
+
+    test "extracts the RTP metadata correctly from packet", %{base_store: store} do
+      packet = PacketFactory.sample_packet(@next_index)
+      {:ok, %PacketStore{heap: heap}} = PacketStore.insert_packet(store, packet)
+
+      assert %Record{index: read_index} = Heap.root(heap)
+
+      assert read_index == @next_index
```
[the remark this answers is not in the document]
sgfn: Removed the test

test/ex_webrtc/rtp/jitter_buffer/realtime_test.exs
```diff
@@ -0,0 +1,117 @@
...
+          } = state
+        ) do
+      now = System.monotonic_time(:millisecond)
+
+      1..packet_num
+      |> Enum.each(fn n ->
+        time =
+          cond do
+            # Delay less than max latency
+            rem(n, 15) == 0 -> n * delay_ms + div(max_latency, 2)
```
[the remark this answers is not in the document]
sgfn: This block of code schedules sending *all* of the packets.

`n * delay_ms` is the base time when the packet is expected to arrive (when it would be sent, were there no delay), while `+ div(max_latency, 2)` is the delay

test/ex_webrtc/rtp/jitter_buffer_test.exs
sgfn: This also asserts that the `PacketStore` was reset (`*_index` set to `nil`, ROC set to `0`...)

test/ex_webrtc/rtp/jitter_buffer_test.exs
```diff
@@ -0,0 +1,127 @@
...
+      flush_index = @base_seq_number - 1
+
+      store = %PacketStore{
+        buffer.store
+        | flush_index: flush_index,
+          highest_incoming_index: flush_index
+      }
+
+      {:ok, store} = PacketStore.insert_packet(store, second_packet)
+      {:ok, store} = PacketStore.insert_packet(store, third_packet)
```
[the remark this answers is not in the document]
sgfn: We want to simulate the case where some packets were placed in the buffer beforehand, then insert one late packet. We avoid any side effects this way, testing this particular case only and not any other related behaviour

Though, if you insist on getting rid of PacketStore in the entire module, I can make that change

test/ex_webrtc/rtp/jitter_buffer_test.exs
sgfn: Changed a few of the calls to `PacketStore`, but ultimately I'm unsure if this is the way to go

sgfn (commented):
First part of second batch of fixes. Will replace the PQ and fix the commented out parts in packet_store_test.exs

lib/ex_webrtc/rtp/jitter_buffer.ex
```diff
@@ -0,0 +1,166 @@
...
+  * `buffer` - `t:#{inspect(__MODULE__)}.t/0`.
+
+  Generally speaking, all results of this type can be handled in the same way.
+  """
+  @type result :: {packets :: [Packet.t()], timer_duration_ms :: timer(), buffer :: t()}
+
+  @opaque t :: %__MODULE__{
+            latency: non_neg_integer(),
+            store: PacketStore.t(),
+            state: :initial_wait | :timer_set | :timer_not_set
```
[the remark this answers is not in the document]
mickel8: > Not differentiating between having the timer set and unset means we'll be telling the user to set a timer every time a packet is inserted

To be precise, every time when usser inserts something and we are waiting for some packet. In most cases, when packets are ordered, we will immediately flush them. 

Anyway, this indeed might not be the bes UX, I didn't notice that earlier :thinking: 

On the other hand, right now, if someone calls handle_timeout too early, they will get `nil` as timer and will think everything is okay

test/ex_webrtc/rtp/jitter_buffer_test.exs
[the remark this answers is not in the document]
mickel8: Please make :pray:

test/ex_webrtc/rtp/jitter_buffer_test.exs
mickel8: The idea is to test JitterBuffer API. In other case, any change in PacketStore API will cause these tests to fail. Sometimes, indeed, we want to check some internal field but we should avoid that as much as possible. In the case of JitterBuffer, I think that we can always find a sequence of packets that after insertion into JitterBuffer  will set PacketStore state in a desired way.

lib/ex_webrtc/rtp/jitter_buffer/packet_store.ex
```diff
@@ -106,86 +106,52 @@ defmodule ExWebRTC.RTP.JitterBuffer.PacketStore do
...
+    packets
   end
 
   @doc """
   Returns timestamp (time of insertion) of the packet with the lowest index
   """
-  @spec first_record_timestamp(t()) :: integer() | nil
-  def first_record_timestamp(%__MODULE__{heap: heap}) do
+  @spec first_entry_timestamp(t()) :: integer() | nil
+  def first_entry_timestamp(%__MODULE__{heap: heap}) do
```
mickel8, line 152:
```suggestion
  def first_packet_timestamp(%__MODULE__{heap: heap}) do
```

lib/ex_webrtc/rtp/jitter_buffer.ex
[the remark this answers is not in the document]
sgfn: >To be precise, every time when usser inserts something and we are waiting for some packet. In most cases, when packets are ordered, we will immediately flush them.

Yeah, you're correct, my bad

>On the other hand, right now, if someone calls handle_timeout too early, they will get nil as timer and will think everything is okay

I don't see how that would happen, we're changing the buffer state to `:timer_not_set` immediately after `handle_timeout` is called, so all of the timer-setting code is being executed
```elixir
  def handle_timeout(buffer) do
    %__MODULE__{buffer | state: :timer_not_set} |> send_packets()
  end
```

test/ex_webrtc/rtp/jitter_buffer/heap_test.exs
```diff
@@ -0,0 +1,74 @@
+defmodule ExWebRTC.RTP.JitterBuffer.HeapTest do
+  use ExUnit.Case, async: true
+
+  alias ExWebRTC.RTP.JitterBuffer.Heap
+
+  test "stores things and reports size" do
```
mickel8, line 6: Nitpick, in Elixir WebRTC we decided to avoid the old Membrane style of describing tests. Sometimes you can still find such tests or it might not be posssible to describe them in other way but in general we favor something like this:

```
test "Heap.push/1"
test "Heap.root/1"
test "Heap.new/1"
```

That's how tests in Elixir repo are written. We don't have to change these tests, just for the future

lib/ex_webrtc/rtp/jitter_buffer/heap.ex
```diff
@@ -0,0 +1,118 @@
+defmodule ExWebRTC.RTP.JitterBuffer.Heap do
```
mickel8, line 1: Awesome, thanks! :tada:

test/ex_webrtc/rtp/jitter_buffer/packet_store_test.exs
```diff
@@ -0,0 +1,282 @@
...
+      #   |> Enum.map(& &1.packet.sequence_number)
+
+      # assert seq_numbers == [65_535, 0, 1]
+
+      # indexes =
+      #   store
+      #   |> PacketStore.dump()
+      #   |> Enum.map(& &1.index)
+
+      # assert indexes == [@seq_number_limit - 1, @seq_number_limit, @seq_number_limit + 1]
```
mickel8, line 107: :eyes:

mickel8 (approved):
Just take care of commented code in tests and merge :+1:

PR #160, opened 2024-08-22
Buffer RT(C)P packets that arrive pre-DTLS handshake completion

sgfn:
This is exceedingly rare but can happen in extremely poor network conditions due to jitter -- and when it happens, it results in dropping the initial keyframe

### DIFF
--- a/lib/ex_webrtc/dtls_transport.ex
+++ b/lib/ex_webrtc/dtls_transport.ex
@@ -362,6 +362,7 @@ defmodule ExWebRTC.DTLSTransport do
         {:ok, state}
 
       {:error, reason} = error ->
+        # TODO: consider buffering DTLS packets that came out of order during the handshake
         Logger.debug("DTLS error: #{reason}")
         error
     end
@@ -393,7 +394,7 @@ defmodule ExWebRTC.DTLSTransport do
     They will be processed after the completion of the handshake.\
     """)
 
-    state = %{state | buffered_remote_rtp_packets: state.buffered_remote_rtp_packets ++ [data]}
+    state = %{state | buffered_remote_rtp_packets: [data | state.buffered_remote_rtp_packets]}
     {:ok, state}
   end
 
@@ -438,10 +439,8 @@ defmodule ExWebRTC.DTLSTransport do
     %{state | remote_cert: cert, remote_base64_cert: base64_cert, remote_fingerprint: fingerprint}
   end
 
-  defp flush_buffered_remote_rtp_packets(%{buffered_remote_rtp_packets: []} = state), do: state
-
   defp flush_buffered_remote_rtp_packets(state) do
-    for data <- state.buffered_remote_rtp_packets do
+    for data <- Enum.reverse(state.buffered_remote_rtp_packets) do
       send(self(), {:ex_ice, state.ice_pid, {:data, data}})
     end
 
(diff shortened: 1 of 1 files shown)


lib/ex_webrtc/dtls_transport.ex
```diff
@@ -433,5 +438,15 @@ defmodule ExWebRTC.DTLSTransport do
     %{state | remote_cert: cert, remote_base64_cert: base64_cert, remote_fingerprint: fingerprint}
   end
 
+  defp flush_buffered_remote_rtp_packets(%{buffered_remote_rtp_packets: []} = state), do: state
```
LVala, line 441: This clause is unnecessary. With an empty list, the `for` will work as expected.

lib/ex_webrtc/dtls_transport.ex
```diff
@@ -384,11 +387,13 @@ defmodule ExWebRTC.DTLSTransport do
...
-    Logger.warning(
-      "Received RTP/RTCP packets, but DTLS handshake hasn't been finished yet. Ignoring."
-    )
+  defp handle_ice_data({:data, data}, state) do
+    Logger.debug("""
+    Received RTP/RTCP packets, but DTLS handshake hasn't been finished yet. \
+    They will be processed after the completion of the handshake.\
+    """)
 
+    state = %{state | buffered_remote_rtp_packets: state.buffered_remote_rtp_packets ++ [data]}
```
LVala, line 396: You could also append the packets `[data | state.buffered_remote_rtp_packets]` and then do `Enum.reverse` in `flush_buffered_remote_rtp_packets`. Probably negligible in terms of performance, but more idiomatic imho.

lib/ex_webrtc/dtls_transport.ex
```diff
@@ -108,6 +108,7 @@ defmodule ExWebRTC.DTLSTransport do
       ice_connected: false,
       buffered_local_packets: nil,
       buffered_remote_packets: nil,
+      buffered_remote_rtp_packets: [],
```
LVala, line 111: We aren't sure if these are RTP packets, it might as well be RTCP (or even SCTP data if we account for it, see other comment), so the name is a bit misleading.

I was thinking of reusing the `buffered_remote_packets`, as it's used for packets before negotiation, then is flushed (affair), so it could be used for that, although this might make the code a bit more cryptic.

LVala (approved):
For now it's probably fine, but we should also consider buffering DTLS packets that came out of order during the handshake (I assume that `ExDTLS` does not take care of that, so we would need to do something in `{:error, reason}` clause in the big `case ExDTLS.handle_data(state.dtls, data) do` statement.

Another thing I was thinking of is buffering packets send by us, because as of now we just throw a log and discard the packets.

If you don't want to take care of these and split attention between this and jitter buffer, just add a comment, especially about the DTLS packets, and I'll take care of that next week.

lib/ex_webrtc/dtls_transport.ex
sgfn: Yeah, I was wary of reusing the same buffer because of that. If you have suggestions for alternative names, I'm all ears

sgfn:
@LVala added a TODO comment in the place you mentioned

lib/ex_webrtc/dtls_transport.ex
LVala: For now leave it, I'll think about it when taking care of buffering DTLS data.

PR #167, opened 2024-09-21
Add support for Fly.io deployments

mickel8:


lib/ex_webrtc/ice/fly_ip_filter.ex
```diff
@@ -0,0 +1,22 @@
+defmodule ExWebRTC.ICE.FlyIpFilter do
+  @moduledoc """
+  ICE IP filter for Fly.io deployments.
+
+  This module defines a single function, which filters out IP addresses,
+  which ICE Agent will use as its host candidates.
```
sgfn, line 6: 1. "filters out" to me sounds like it's rejecting these addresses rather than accepting them
2. I feel like adding a bit more info about the function could be helpful

```suggestion
  This module defines a single function, which filters IP addresses,
  which ICE Agent will use as its host candidates.
  It accepts only the IPv4 address that `fly-global-services` resolves to.
```

lib/ex_webrtc/ice/fly_ip_filter.ex
```diff
@@ -0,0 +1,22 @@
...
+  """
+
+  @spec ip_filter(:inet.ip_address()) :: boolean()
```
sgfn, line 9: [nitpick] We could add a behaviour in ExICE. Then again, it's only one function -- up to you

guides/deploying/fly.md
```diff
@@ -0,0 +1,50 @@
...
+
+In JavaScript code:
+
+```js
+pc = new RTCPeerConnection({
+  iceServers: [{ urls: "stun:stun.l.google.com:19302" }],
+});
+```
+
+in Elixir code:
```
sgfn, line 20:
```suggestion
In Elixir code:
```

guides/deploying/fly.md
```diff
@@ -0,0 +1,50 @@
...
+ip_filter = Application.get_env(:your_app, :ice_ip_filter)
+
+{:ok, pc} =
+  PeerConnection.start_link(
+    ice_ip_filter: ip_filter,
+    ice_servers: [%{urls: "stun:stun.l.google.com:19302"}]
+  )
+```
+
+in `runtime.exs`:
```
sgfn, line 32:
```suggestion
In `runtime.exs`:
```

guides/deploying/fly.md
```diff
@@ -0,0 +1,50 @@
...
+
+```elixir
+if System.get_env("FLY_IO") do
+  config :your_app, ice_ip_filter: &ExWebRTC.ICE.FlyIpFilter.ip_filter/1
+end
+```
+
+in fly.toml:
```
sgfn, line 40:
```suggestion
In fly.toml:
```

guides/deploying/fly.md
```diff
@@ -0,0 +1,50 @@
...
+
+```elixir
+ip_filter = Application.get_env(:your_app, :ice_ip_filter)
+
+{:ok, pc} =
+  PeerConnection.start_link(
+    ice_ip_filter: ip_filter,
+    ice_servers: [%{urls: "stun:stun.l.google.com:19302"}]
+  )
+```
```
sgfn, line 30: [nitpick] When `:ice_ip_filter` is unset in the application, we're passing `ice_ip_filter: nil` to `PeerConnection`, which means https://github.com/elixir-webrtc/ex_webrtc/blob/e42c09df1ae702126dc407ee35a5760d0f601967/lib/ex_webrtc/peer_connection/configuration.ex#L242 

doesn't insert the default filter.

When we pass it to ICE, however, this line
https://github.com/elixir-webrtc/ex_ice/blob/4c5190687e11b28a8209f2c930895647d0b2a64c/lib/ex_ice/priv/ice_agent.ex#L137
makes it work again, as it gets the `nil` stored in the keyword list and overrides it with the default filter.

I feel like this creates a kind of ambiguity and depends on ExICE not changing its behaviour of parsing options, which can potentially lead to hard-to-find errors in the future. To avoid this in this example, I would explicitly *not* insert `ice_ip_filter` to the PC options when it's missing from the application's environment.

guides/deploying/fly.md
```diff
@@ -0,0 +1,50 @@
...
+in fly.toml:
+
+```toml
+[env]
+  # add one additional env
+  FLY_IO = 'true'
+```
+
+That's it!
+No special UDP port exports or dedicated IP address needed.
```
sgfn, line 49:
```suggestion
No special UDP port exports or dedicated IP addresses are needed.
```

lib/ex_webrtc/ice/fly_ip_filter.ex
mickel8: Let's stay with the current version for know. In debugging and testing it's verry convenient to pass anonymous function instead of creating a module with a single function.

I thought about returning an anonymous function from `ip_filter` and use `ICEAgent.ip_filter` type in typespec but sounds like an overkill :thinking:

guides/deploying/fly.md
mickel8: Good catch!

sgfn (approved):

PR #168, opened 2024-09-23
Update deps and examples

mickel8:
This PR:
* update deps in ex_webrtc and example apps
* handles :EXIT message from PeerConnection - Bandit seems to trap exits so when our PeerConnection goes down, it won't bring PeerHandler process down. In most cases, PeerHandler is terminated anyway as WS closes on the other hand but...
* fixed two obsoleted functions from JB


examples/save_to_file/lib/save_to_file/peer_handler.ex
```diff
@@ -83,13 +83,21 @@ defmodule SaveToFile.PeerHandler do
       :video -> state.video_buffer
       :audio -> state.audio_buffer
     end
-    |> JitterBuffer.handle_timer()
+    |> JitterBuffer.handle_timeout()
```
sgfn, line 86: oops, sorry bout that

sgfn (approved):

PR #170, opened 2024-10-08
Fix rtcp for no track

roznawsk:
It may happen, that after `removing_track`, we still receive PLI for that track, which has been already removed from the transceiver.

Below are logs from the fishjam media server

[pli-fail-logs.txt](https://github.com/user-attachments/files/17290362/pli-fail-logs.txt)

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -2087,6 +2087,9 @@ defmodule ExWebRTC.PeerConnection do
       nil ->
         {nil, state}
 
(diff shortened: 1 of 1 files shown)


mickel8 (commented):
We should probably do the same for all other `handle_rtcp_packet` clauses :thinking:

roznawsk:
I added it in the other case, where we handle rtcp on an outbound track. Let me know if you think the issue can appear also on the inbound tracks.

mickel8:
Thanks!

mickel8:
Released as 0.6.1

PR #176, opened 2024-11-28
Use track id in get_stats

roznawsk:


mickel8 (changes requested):
Track attached to the sender can be changed without creating a new sender (using `replaceTrack` function). I think we can store track id in a new field called  `mediaSourceId`, see [here](https://www.w3.org/TR/webrtc-stats/#dom-rtcoutboundrtpstreamstats-mediasourceid)

Could you also add [`trackIdentifier`](https://www.w3.org/TR/webrtc-stats/#dom-rtcinboundrtpstreamstats-trackidentifier) to the receiver stats?

mickel8 (approved):

PR #180, opened 2024-12-18
Track recordings

sgfn:


lib/ex_webrtc/recorder.ex
```diff
@@ -0,0 +1,193 @@
...
+  @spec add_tracks(GenServer.server(), [MediaStreamTrack.t()]) :: :ok | :error
+  def add_tracks(recorder, tracks) do
+    # XXX need?
+    try do
+      GenServer.call(recorder, {:add_tracks, tracks})
+    catch
+      _exit_or_error, _e ->
+        Logger.error("Recorder is down, not adding tracks")
+        :error
+    end
```
sgfn, line 72: Up for discussion whether this `try` block should go here

I don't think this is the correct place for it, I'd put it in Broadcaster logic instead

mickel8: I had the same problem some time ago - [link](https://elixirforum.com/t/how-to-correctly-handle-genserver-call-exits-because-of-non-existing-process/58284) :roll_eyes: 

IMO, we can remove `try catch` for now. I don't know how someone could handle the error as it is related to this specific track, not the process of recording as a whole. 

Recorder is also assumed to be alive all the time so if it is not, then something bad happened and it might be better for the caller to just crash. 

And we can always go back to `try catch` if we find it useful in the future.

lib/ex_webrtc/recorder/converter.ex
```diff
@@ -0,0 +1,171 @@
...
+        Map.new(stores, fn {rid_idx, store} ->
+          {rid_idx, store |> PacketStore.dump() |> Enum.reject(&is_nil/1)}
+        end)
+    end
+  end
+
+  defp read_packet(file) do
+    case IO.binread(file, 13) do
+      <<rid_idx::8, _recv_time::64, packet_size::32>> ->
+        with {:ok, packet} <- file |> IO.binread(packet_size) |> ExRTP.Packet.decode() do
```
mickel8, line 152: This IO.binread can fail too

lib/ex_webrtc/recorder/converter.ex
```diff
@@ -0,0 +1,171 @@
...
+        "video" ->
+          convert_video_track(id, rid_map, output_path, packets)
+
+        "audio" ->
+          {:ok, writer} =
+            output_path
+            |> Path.join("#{id}.ogg")
+            |> Ogg.Writer.open()
+
+          {:ok, depayloader} = Depayloader.new(@audio_codec_params)
```
mickel8, line 87: Nitpick: I would wrap this into `convert_audio_track`

lib/ex_webrtc/recorder/converter.ex
```diff
@@ -0,0 +1,171 @@
...
+        :eof
+    end
+  end
+
+  defp insert_packet_to_store(store, packet) do
+    case PacketStore.insert(store, packet) do
+      {:ok, store} ->
+        store
+
+      {:error, :late_packet} ->
```
mickel8, line 166: In practice, this should only happen when a recording starts with out-of-order packets, correct?

lib/ex_webrtc/recorder/converter.ex
```diff
@@ -0,0 +1,171 @@
...
+  def convert!(report_path, output_path \\ @default_output_path) do
+    report_path =
+      report_path
+      |> Path.expand()
+      |> then(
+        &if(File.dir?(&1),
+          do: Path.join(&1, "report.json"),
+          else: &1
+        )
+      )
```
mickel8, line 55: I think we should take a path to the base dir and process all subdirectories. Not necessarily in this PR but in general. Optionally, we can allow to pass an option to process a single dir.

mickel8 (commented):
Really nice and concise code! 

There are two blocking comments - about using time for dir names, and taking base dir path as an argument in the converter

lib/ex_webrtc/recorder/converter.ex
sgfn: Yes, I believe that's correct

lib/ex_webrtc/recorder.ex
```diff
@@ -0,0 +1,193 @@
...
+  def record(recorder, track_id, rid, %ExRTP.Packet{} = packet) do
+    recv_time = System.monotonic_time(:millisecond)
+    GenServer.cast(recorder, {:record, track_id, rid, recv_time, packet})
+  end
+
+  @impl true
+  def init(config) do
+    base_dir =
+      (config[:base_dir] || @default_base_dir)
+      |> Path.join(UUID.uuid4())
```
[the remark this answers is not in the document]
sgfn: No strong feelings either way. I'll make that change

lib/ex_webrtc/recorder/converter.ex
sgfn: I agree, but I'd leave it for when we need to use the Converter somewhere (e.g in our demo)

mickel8 (approved):

PR #181, opened 2024-12-29
Handle ICE candidates with missing `usernameFragment`

samrat:
I ran into an error because `flutter-webrtc`(https://github.com/flutter-webrtc/webrtc-interface/blob/767eb083a9ee88a4e37b49cb70bb7831c7c2414d/lib/src/rtc_ice_candidate.dart#L6) doesn't seem to send this field.

The ICECandidate type spec already allows the field to be nil. This PR would also allow candidates where the `usernameFragment` field is not present at all.

### DIFF
--- a/lib/ex_webrtc/ice_candidate.ex
+++ b/lib/ex_webrtc/ice_candidate.ex
@@ -23,6 +23,16 @@ defmodule ExWebRTC.ICECandidate do
   end
 
   @spec from_json(%{String.t() => String.t() | non_neg_integer() | nil}) :: t()
+  def from_json(
+        %{
+          "candidate" => c,
+          "sdpMid" => mid,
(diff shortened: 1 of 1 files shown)


mickel8:
Thanks! I confirmed that `usernameFragment` indeed might be null according to [RTCIceCandidateInit](https://www.w3.org/TR/webrtc/#dictionary-rtcicecandidateinit-members) interface and procedure describing the creation of RTCIceCandidate object - [link](https://www.w3.org/TR/webrtc/#constructor-0)

mickel8:
I slightly refactored the code in https://github.com/elixir-webrtc/ex_webrtc/commit/117f39ec4795aa5871f2a201b6cdb386e492dd94

PR #185, opened 2025-01-15
Add ssrc attributes to the SDP

mickel8:
Some implementations still rely on ssrc attributes in SDP offer/answer when demuxing incoming RTP packets (e.g. Pion). While including a=ssrc attribute in the SDP is obsoleted and we should demux by mid, including ssrc is still correct and supported by web browsers.

This PR adds ssrc attributes to the SDP offer/answer when a transceiver is sendonly or sendrecv.

A couple of notes:

* in theory, if we add `a=ssrc:1234` attribute, it should also always has cname attr i.e. `a=ssrc:1234 cname:qW34eced` (RFC 5567)
* we don't use cnames so we don't do this. Instead we always add msid attribute as we already have it
* we stop sending app_data in msid according to what JSEP says (RFC 8829). This app_data does not make any sense anyway as its value is MediaStreamTrack.id, which can change without renegotiation (e.g. when calling rtpsender.replaceTrack)
* we preallocate ssrc and rtx_ssrc even for recvonly transceivers. This is not fully correct, as ssrc should probably change after transceiver becomes inactive and active again but as far as I understand the code, we haven't been doing this so far anyway
* we don't add ssrc, ssrcgroup and msid attributes when transceiver is recvonly or inactive
* relevant rfcs: 
  * [RFC 8830](https://www.rfc-editor.org/rfc/rfc8830.html) - MediaStreams 
  * [RFC 8829](https://datatracker.ietf.org/doc/html/rfc8829) - JSEP 
  * [RFC 5576](https://datatracker.ietf.org/doc/html/rfc5576) - SSRC attribute and FID group

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -870,7 +870,6 @@ defmodule ExWebRTC.PeerConnection do
   @impl true
   def handle_call({:add_track, %MediaStreamTrack{kind: kind} = track}, _from, state) do
     # we ignore the condition that sender has never been used to send
-    {ssrc, rtx_ssrc} = generate_ssrcs(state)
 
     {transceivers, sender} =
       state.transceivers
@@ -878,10 +877,12 @@ defmodule ExWebRTC.PeerConnection do
       |> Enum.find(fn {tr, _idx} -> RTPTransceiver.can_add_track?(tr, kind) end)
       |> case do
         {tr, idx} ->
-          tr = RTPTransceiver.add_track(tr, track, ssrc, rtx_ssrc)
+          tr = RTPTransceiver.add_track(tr, track)
           {List.replace_at(state.transceivers, idx, tr), tr.sender}
 
         nil ->
+          {ssrc, rtx_ssrc} = generate_ssrcs(state)
+
           options = [
             direction: :sendrecv,
             added_by_add_track: true,
@@ -910,8 +911,7 @@ defmodule ExWebRTC.PeerConnection do
         {:reply, {:error, :invalid_track_type}, state}
 
       {tr, idx} when tr.direction in [:sendrecv, :sendonly] ->
-        {ssrc, rtx_ssrc} = generate_ssrcs(state)
-        tr = RTPTransceiver.replace_track(tr, track, ssrc, rtx_ssrc)
+        tr = RTPTransceiver.replace_track(tr, track)
         transceivers = List.replace_at(state.transceivers, idx, tr)
         state = %{state | transceivers: transceivers}
         {:reply, :ok, state}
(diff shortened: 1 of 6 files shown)


lib/ex_webrtc/peer_connection.ex
```diff
@@ -1835,25 +1841,40 @@ defmodule ExWebRTC.PeerConnection do
...
         do: :inactive,
         else: SDPUtils.get_media_direction(mline) |> reverse_direction()
 
+    rtp_sender_ssrcs = Enum.map(transceivers, & &1.sender.ssrc)
+    ssrcs = MapSet.new(demuxer_ssrcs ++ rtp_sender_ssrcs)
+
+    ssrc = do_generate_ssrc(ssrcs)
+    ssrcs = MapSet.put(ssrcs, ssrc)
+    rtx_ssrc = do_generate_ssrc(ssrcs)
+
```
sgfn, line 1871: Would it be a huge pain to rewrite this function so that it accepts `state` and uses `generate_ssrcs`, where this logic is present already?

test/ex_webrtc/rtp_transceiver_test.exs
```diff
@@ -0,0 +1,85 @@
+defmodule ExWebRTC.RTPTransceiverTest do
+  use ExUnit.Case, async: true
+
+  alias ExWebRTC.{MediaStreamTrack, PeerConnection, RTPTransceiver, Utils}
+
+  {:ok, pc} = PeerConnection.start_link()
+  @config PeerConnection.get_configuration(pc)
+  :ok = PeerConnection.close(pc)
```
sgfn, line 8: O_O is this even legal

sgfn (commented):
As far as I understand the logic, everything looks good 👍 👍 👍

lib/ex_webrtc/rtp_sender.ex
```diff
@@ -131,6 +134,78 @@ defmodule ExWebRTC.RTPSender do
...
+        {ssrc_attrs, rtx_ssrc_attrs} =
+          Enum.reduce(streams, {[], []}, fn stream, {ssrc_attrs, rtx_ssrc_attrs} ->
+            ssrc_attr = %ExSDP.Attribute.SSRC{id: ssrc, attribute: "msid", value: stream}
+            ssrc_attrs = [ssrc_attr | ssrc_attrs]
+
+            rtx_ssrc_attr = %ExSDP.Attribute.SSRC{id: rtx_ssrc, attribute: "msid", value: stream}
+            rtx_ssrc_attrs = [rtx_ssrc_attr | rtx_ssrc_attrs]
+
+            {ssrc_attrs, rtx_ssrc_attrs}
+          end)
```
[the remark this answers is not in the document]
mickel8: As this function is not so long and it only does a single thing (generates only ssrc attrs) I would leave it as is. If it do multiple things like generating ssrc attributes, generate msid, etc. I would also opt for splitting generation of every attribute into a private function

test/ex_webrtc/rtp_transceiver_test.exs
mickel8: Yeap, but acutally I am curious what is the process, peer connection will be linked to :thinking:

lib/ex_webrtc/peer_connection.ex
mickel8: Good idea :+1:

mickel8: After more detailed analysis, I would prefer this function to return transceivers as this gives us more flexibility (instead of taking state and returning a new state). We then call different functions like remove_stopped_transceivers and in general the flow is pretty complex so I feel like this function shouldn't modify the state. What I can do is to modify generate_ssrcs to take demuxer and transceivers

sgfn (approved):

PR #189, opened 2025-01-30
Fix RC in data_channel_test.exs

mickel8:
`PeerConnection.send_data` is asynchronous, so before checking stats, we need to ensure that the other side received our messages.


sgfn (approved):
<3

PR #190, opened 2025-01-30
Add set_sender_codec

mickel8:
This PR allows to set codec that should be used for sending with the following remarks:
* when the first RTP packet is sent, the clock rate used by RTP sender is "locked" meaning that any subsequent call to the `set_sender_codec` must use a codec with the same clock rate. This is because using a new clock rate requires a new SSRC. Also, ReportRecorder, which generates RTCP reports relies on clock rate so changing it after we started sending packets could result in incorrect reports
* when a codec was selected, and as a result of renegotiation it is no longer supported, user must call `set_sender_codec` with a new codec that is supported

It also:
* moves initialization of ReportRecorder so that now, it is initialized when the first packet is sent
* adds checking FMTP parameters  when negotiating codecs (if codec's sdp_fmtp_line is not nil, otherwise sdp_fmtp_line is always assumed to be the same)
* ensures that codec payload types and rtp header extension ids are not changed when they are not used by the remote description (we have been updating them so far even when there was no need to do that)
* sets RTPSender codecs after negotiation (instead of during creation time)

RFC considerations:
* Codecs listed in the remote description are in the preference order i.e. the first codec is the most preferred to receive by the remote side, and we as a sender SHOULD use it for sending.
* However, RFC 3264 (sec 6.1 and 7.0) allows for changing codec on the fly (e.g. during silence periods although Opus solves this problem and I have never seen anyone doing this) and receiver MUST be prepared to handle this situation
* Relying on these requirements, we allow user to set the codec used for sending

### DIFF
--- a/lib/ex_webrtc/rtp_sender.ex
+++ b/lib/ex_webrtc/rtp_sender.ex
@@ -263,8 +263,8 @@ defmodule ExWebRTC.RTPSender do
   defp rtx?(codec), do: String.ends_with?(codec.mime_type, "rtx")
   defp supported?(sender, codec), do: codec in sender.codecs
 
-  # As long as report recorder is not initialized i.e. we have not send any RTP packet
-  # allow for codec changes. Once we start sending RTP packet, require the same clock rate.
+  # As long as report recorder is not initialized i.e. we have not sent any RTP packet,
+  # allow for codec changes. Once we start sending RTP packets, require the same clock rate.
   defp same_clock_rate?(%{report_recorder: %{clock_rate: nil}}, _codec), do: true
   defp same_clock_rate?(sender, codec), do: sender.report_recorder.clock_rate == codec.clock_rate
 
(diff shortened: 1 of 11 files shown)


lib/ex_webrtc/rtp_sender.ex
```diff
@@ -118,32 +121,56 @@ defmodule ExWebRTC.RTPSender do
...
+    log_rtx_codec_change(sender, rtx_codec, codecs)
+
     # TODO: handle cases when codec == nil (no valid codecs after negotiation)
     pt = if codec != nil, do: codec.payload_type, else: nil
     rtx_pt = if rtx_codec != nil, do: rtx_codec.payload_type, else: nil
 
-    report_recorder = %ReportRecorder{
-      sender.report_recorder
-      | clock_rate: codec && codec.clock_rate
-    }
```
mickel8, line 133: Clock rate of ReportRecorder shouldn't change as it may result in incorrect RTCP reports. Instead, we probably should generate a new ssrc and start sending new RTCP reports. Anyways, we don't allow for codec change that has different clock rate so we can safely delete this code

PR #194, opened 2025-02-11
feat: Merge tracks/refactor do_add_tracks

jongirard:
Consider the following utilization of `ex_webrtc` recorder (as I am using in my current project):

```
- Room module starts and manages a single recorder process per room via RecorderSupervisor, preventing duplicate recorders
- Peers register their media tracks with the recorder when tracks are created via the :track event handler
- Peers send RTP packets to the recorder as they receive them via the :rtp event handler
- (This separation keeps recording coordination in Room while Peers focus on media handling)
```

The problem is the current recorder implementation in `do_add_tracks` is replacing the existing tracks map and returning a new map anytime we call `add_tracks` rather than merging the new tracks with existing ones. This causes `[warning] Tried to save packet for unknown track id. Ignoring. Track id: xxxx` warnings and failed recordings.

This PR refactors the `do_add_tracks` functionality and merges existing tracks with new tracks thereby fixing the above mentioned problem. Another option would be to introduce a new function called something like `do_add_merged_tracks` instead.

### DIFF
--- a/lib/ex_webrtc/recorder.ex
+++ b/lib/ex_webrtc/recorder.ex
@@ -153,31 +153,70 @@ defmodule ExWebRTC.Recorder do
     {:noreply, state}
   end
 
-  defp do_add_tracks(tracks, state) do
+  defp do_add_tracks(tracks, %{base_dir: base_dir, tracks: existing_tracks} = state) do
     start_time = DateTime.utc_now()
 
-    tracks =
-      Map.new(tracks, fn track ->
-        path = Path.join(state.base_dir, "#{track.id}.rtpx")
-        file = File.open!(path, [:write])
-        rid_map = (track.rids || [nil]) |> Enum.with_index() |> Map.new()
+    # Create new track entries with computed metadata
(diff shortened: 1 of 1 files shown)


sgfn:
Hi @jongirard, thanks for the bug report and the fix, much appreciated.

It seems we were erroneously updating the tracks in `state` for a second time in the final line of `do_add_tracks/2`, so the adjustment from #195 should be enough to fix the issue you've encountered.

Thank you also for the additional refactor you proposed. However, the implementation of the `Recorder` ecosystem is currently in a delicate phase as we are actively working on it and introducing new functionalities. Because of these ongoing changes and our current bandwidth limitations, we're unable to merge broader changes at this moment.

For the time being, we'll proceed with the exact fix mentioned above. Could you please test it and confirm if it resolves the problem on your end?

jongirard:
@sgfn thanks for the proposed solution, I tested the fix today in my project and #195 does indeed resolve the problem! Closing this PR.

PR #196, opened 2025-02-12
feat: Optional filename prefix for tracks

jongirard:
As @sgfn mentioned over in #194, I'm aware the recorder ecosystem is in a delicate state right now with limited bandwidth for broader changes, however, I currently have a need within my project to support a known identifier in created recordings (which I believe may be a use case others will encounter as well).

This minor change will allow for an optional prefix to be used in a track's filename (such as user id) which is useful for associating the arbitrary track ID with a known identifier (ex: `e0bc96adf4-19760930259463780812500431447.rtpx` (user_id-track_id).

Thank you for considering this change!

### DIFF
--- a/lib/ex_webrtc/recorder.ex
+++ b/lib/ex_webrtc/recorder.ex
@@ -60,9 +60,9 @@ defmodule ExWebRTC.Recorder do
   @doc """
   Adds new tracks to the recording.
   """
-  @spec add_tracks(GenServer.server(), [MediaStreamTrack.t()]) :: :ok
-  def add_tracks(recorder, tracks) do
-    GenServer.call(recorder, {:add_tracks, tracks})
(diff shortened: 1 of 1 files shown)


sgfn:
Thanks for the suggestion. We'll make sure to add the possibility of including custom metadata together with the recordings in 0.9.0

You can track the progress in #197 :)

jongirard:
@sgfn thanks, ideally the metadata can be _per track_ like this, not simply at the room level. This gives the ultimate flexibility imo. I also personally like having the prefix in the filename too because it's easy to observe at a high level/in a folder hierarchy. Coupled with the option to include a metadata map for any other extra fields in the reports.json which could be nice.

sgfn:
hi, if this issue is still relevant to you, please reopen it in https://github.com/elixir-webrtc/ex_webrtc_recorder as that's where the recorder code was extracted

thanks

PR #197, opened 2025-02-12
Recorder+Converter enhancements

sgfn:
## Features
- Converter now combines video and audio tracks belonging to the same stream, syncs them and outputs a WEBM file
  - When using simulcast, this results in 3 output files, one per layer
- Converter can generate thumbnails for the result files
- Recorder can upload the RTP dumps to S3-compatible storage
- Converter can download the RTP dumps from S3-compatible storage, and upload the conversion results there as well

### Recorder
- Recorder now has a controlling process to which it sends messages
  - At the moment, messages are sent only when using S3 uploads
- The nomenclature of `report` (and the `report.json` file) was changed to `manifest` (`manifest.json`)
- `add_tracks/2` now returns `{:ok, manifest :: Recorder.manifest()}`
  - `manifest` contains only the parts relevant to the added tracks, and not the entire manifest of the Recorder

### Converter
- Converter now generates WEBM files using FFmpeg, so it requires the `ffmpeg` binary with the relevant libraries present in `PATH`
- Removed `convert!/2`. Use `convert_path!/2` instead
  - Note that the new function accepts `t:Converter.options/0` as the second argument

### Recorder
- New function `end_tracks/2` that closes the relevant file handles and optionally schedules the S3 upload

### Converter
- New function `convert_manifest!/2` that can be used together with the manifest returned by `Recorder.end_tracks/2`


lib/ex_webrtc/recorder/converter.ex
```diff
@@ -37,105 +46,324 @@ defmodule ExWebRTC.Recorder.Converter do
...
   @doc """
-  Convert the saved dumps of tracks in the report to IVF and Ogg files.
+  Loads the recording manifest from file, then proceeds with `convert_manifest!/2`.
   """
-  @spec convert!(Path.t(), Path.t()) :: :ok | no_return()
-  def convert!(report_path, output_path \\ @default_output_path) do
-    report_path =
-      report_path
+  @spec convert_path!(Path.t(), options()) :: __MODULE__.Manifest.t() | no_return()
+  def convert_path!(recorder_manifest_path, options \\ []) do
```
mickel8, line 92: Nitpick: I would call this convert and use pattern matching to see whether we got path or manifest as an argument

lib/ex_webrtc/recorder/s3/upload_handler.ex
```diff
@@ -0,0 +1,135 @@
...
+
+      download_manifest =
+        Map.new(manifest, fn {id, object_data} ->
+          {:ok, location} = Recorder.S3.Utils.to_url(bucket_name, s3_paths[id])
+
+          {id, %{object_data | location: location}}
+        end)
+
+      # FIXME: this links, ideally we should spawn a supervised task instead
+      task = Task.async(fn -> upload(manifest, bucket_name, s3_paths, s3_config_overrides) end)
```
mickel8, line 60: You can use application supervision tree

mickel8 (commented):
Great job! :tada: 

I didn't expect this code to expand so much :eyes: 

One thing that comes to my mind is that we should move this to a separate package. Mostly because it brings a lot of optional dependencies and the logic (uploading, downloading to/from S3) is not strictly related to WebRTC. Initially, I thought this will be a small addition but it turned out to be pretty advanced.

Also, I believe this will help us make better API, a new package can have its own supervision tree where uploader tasks will be spawned, and will not affect casual webrtc development (tests duration, compilation time, etc.) 

WDYT?

sgfn:
@mickel8 I agree about the separation

Closing this in favour of #199 and https://github.com/elixir-webrtc/ex_webrtc_recorder/pull/1

PR #200, opened 2025-02-24
feat: Add `data_type` configuration to `PeerConnection.send_data`

SamWolfs:
- Added a parameter to the `PeerConnection.send_data` function to allow for the choice between `:binary` and `:string`.
- Added a simple test to verify the new `PeerConnection.send_data/4` arity.

No tests were added to validate the correctness of the payload, as it's handled by [ex_sctp](https://github.com/elixir-webrtc/ex_sctp) and would require some sort of passthrough mock. Please let me know if this would require testing and what the preferred approach would be.

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -245,9 +245,9 @@ defmodule ExWebRTC.PeerConnection do
 
   If `ref` does not identify any DataChannel, this function behaves like no-op.
   """
-  @spec send_data(peer_connection(), DataChannel.ref(), binary()) :: :ok
-  def send_data(peer_connection, channel_ref, data) do
-    GenServer.cast(peer_connection, {:send_data, channel_ref, data})
+  @spec send_data(peer_connection(), DataChannel.ref(), binary(), :string | :binary) :: :ok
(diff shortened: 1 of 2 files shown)


mickel8:
Hi @SamWolfs!

Thanks for the PR. I wonder what's the use-case for choosing between string and binary? Checkin the official JS API, I can't see this info being exposed in any-way?

SamWolfs:
Hi @mickel8 

We're sending audio with an unsupported media type over the DataChannel to a [GStreamer DataChannel Peer](https://gstreamer.freedesktop.org/documentation/webrtclib/gstwebrtc-datachannel.html), which cares about the payload type.

mickel8:
Yeah, I just wonder what's the difference between string and binary as this option seems not to be exposed in JS API, yet you can still send files (hence binaries). Maybe it always should be a binary :thinking:

mickel8:
Yeah, I also took a look here: https://www.rfc-editor.org/rfc/rfc8831.html#section-5-6

So PPI (Payload Protocol Identifier) is passed by upper layer and then provided to the upper layer on the other side. It's not meant for SCTP itself. What's more in JS API it's not available (I assume they always use binary).

However, if other "backend" WebRTCimplementations care about it, I think we can make it configurable. 

Thanks for the PR!

PR #202, opened 2025-03-03
Don't terminate PeerConnection after moving to the failed state

mickel8:
We have been terminating PeerConnection after it moved to the failed state. While this approach was easy to implement, it had two drawbacks:
* once PeerConnection failed, we couldn't gather stats via `get_stats` as the process was no longer alive
* we couldn't perform ICE restart

**This PR keeps peer connection alive even when it moves to the failed state. This is pretty big breaking change as a lot of our examples and demo apps relies on linking to the PeerConnection process and waiting until it crashes.**

Things to note:
* when PC moves to the failed state (no matter this is because of DTLS or ICE transport fails), most of the operations are still allowed. E.g. user can still successfully call `addTrack`, `addTransceiver` or `createOffer` even though the connection might not be able to restart (that's the case when DTLS fails, or at least we don't support DTLS restart). 
* sending RTP, RTCP, or data via DataChannel in state failed is silently discarded - silently because there is a RC between calling `send_rtp` and others, and moving to the failed state so such a log could appear pretty often
*  closing web browser's peer connection no longer terminates our peer connection process. So far, closing web browser's peer connection was causing DTLS alert (probably close_notify) that was resulting in the termination of DTLS transport process. From now, we just move DTLS transport to the failed state  

Relevant specs:
* W3C WebRTC sec. 5.5 - https://www.w3.org/TR/webrtc/#rtcdtlstransport-interface


lib/ex_webrtc/peer_connection.ex
```diff
@@ -1283,21 +1301,15 @@ defmodule ExWebRTC.PeerConnection do
   @impl true
   def handle_info({:ex_ice, _from, {:connection_state_change, new_ice_state}}, state) do
     state = %{state | ice_state: new_ice_state}
+    notify(state.owner, {:ice_connection_state_change, new_ice_state})
```
mickel8, line 1304: This one should be fired before connection state change

PR #212, opened 2025-04-11
fix order() typespec, mismatch with the sctp

yakiro-nvg:


sgfn:
typespec seems correct to me, closing

PR #217, opened 2025-05-27
get ice-lite returning nil

OctopusRage:
fix get ice lite not recognized sdp offer that has "a=ice-lite"

### DIFF
--- a/lib/ex_webrtc/sdp_utils.ex
+++ b/lib/ex_webrtc/sdp_utils.ex
@@ -140,7 +140,7 @@ defmodule ExWebRTC.SDPUtils do
   end
 
   @spec get_ice_lite(ExSDP.t()) :: boolean()
-  def get_ice_lite(sdp), do: (ExSDP.get_attribute(sdp, "ice-lite") || ExSDP.get_attribute(sdp, :ice_lite)) != nil
+  def get_ice_lite(sdp), do: ExSDP.get_attribute(sdp, :ice_lite) != nil
 
   @spec get_ice_credentials(ExSDP.t()) ::
           {:ok, {binary(), binary()} | nil}
--- a/lib/ex_webrtc/sdp_utils.ex
+++ b/lib/ex_webrtc/sdp_utils.ex
@@ -140,7 +140,7 @@ defmodule ExWebRTC.SDPUtils do
   end
 
   @spec get_ice_lite(ExSDP.t()) :: boolean()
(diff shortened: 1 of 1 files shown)


mickel8:
Hi @OctopusRage! Thanks for the catch! I think that we could even get rid of the first getter i.e. always look for `:ice_lite`, not for `"ice-lite"`. Could you do that as well?

mickel8:
And see if it works for you?

OctopusRage:
sure, updated

mickel8:
Thanks! Btw. can I ask what's your use-case for ice-lite?

mickel8:
But even without this fix, everything should work, I believe? The worst case scenario is that there will be ICE role conflict that should be automatically resolved

mickel8:
close notify is an alert sent by a side that is closing the connection in a graceful way. When you close a tab in a web browser it should also send DTLS close notify. So this error might not tell anything specific. See here: https://datatracker.ietf.org/doc/html/rfc5246#section-7.2.1

Can you confirm that you are receiving any packets? Also, are you sure you are sending any packets? Some services may close the connection when they don't receive any traffic.

OctopusRage:
Youre right, they immediately close connection after 20 sec if thereis no packet sent
I think i alr send packet to others peers using PeerConnection.send_rtp
I managed to estabilished call and i can hear it both on my browser and my mobile, but it still closed after 1 minutes of call.

Do i probably need to setup turn server on my local as well?

OctopusRage:
i dont think so, all http request done before call estabilished
maybe you can  figured this out @mickel8 ?

image
on my ss above, i started to get get rtcp msg `{nil,   %ExRTCP.Packet.SourceDescription{}` then as soon as im getting those msg i always get TLS error
does it mean i disconnected from their side?

OctopusRage:
i guess ill try to ask meta team too

mickel8:
You can check against `{:ex_webrtc, _pid, {:connection_state_change, conn_state}`, this is the message emitted by ex_webrtc. Also, you can try to enable debug logs and see if there are any ICE-related errors

OctopusRage:
Hi @mickel8 sry to bring this up again. I just revisited my code and i figured it out.
My peer connection suddenly stopped sending to their server after 1 mins. What might be causing this?
My ice connection is fine and connection state change is complete / connected and i still receiving their packet.

PR #220, opened 2025-06-19
Look for SPS instead of keyframe with H264 packetization_mode=0

sgfn:
As explained in the comments, checking for `nalu_type == 7` with H264 packetization_mode=0 has been verified to work correctly under simulated packet loss conditions.


lib/ex_webrtc/rtp/h264.ex
```diff
@@ -22,7 +28,7 @@ defmodule ExWebRTC.RTP.H264 do
   def keyframe?(%Packet{}), do: false
 
   defp do_keyframe?(0, _), do: false
-  defp do_keyframe?(nalu_type, _) when nalu_type in 1..23, do: nalu_type == 5
+  defp do_keyframe?(nalu_type, _) when nalu_type in 1..23, do: nalu_type == 7
```
mickel8, line 31: Should be the same as:
```suggestion
  defp do_keyframe?(7, _), do: true
```

mickel8 (approved):

lib/ex_webrtc/rtp/h264.ex
sgfn: Yeah, but I believe the explicitness of the current version is better. It communicates clearly that we're considering the case with `packetization_mode=0` (Single NAL Unit Mode), where we have exactly 1 AU in each NALU. I'd leave it as is.

PR #237, opened 2025-11-20
bugfix/dtls-crash-onclose

OctopusRage:
- the strange behaviour occured randomly when my peer tried to close its connection
- this fix dtls_transport = nil causing peer connection crashes when closed
- already try catch `PeerConnection.close()`. but still cannot avoid this error because its nif  implementation (?)
- this pr eliminate my current workaround by checking `PeerConnection.get_dtls_transport_state(peer_connection) not in [:failed, :closed]` before closing peer connection

image


sgfn:
should be fixed by #251

PR #238, opened 2025-11-20
TCP ICE first iteration

sgfn:
ref: https://github.com/elixir-webrtc/ex_ice/pull/89


lib/ex_webrtc/rtp_transceiver.ex
```diff
@@ -196,6 +198,21 @@ defmodule ExWebRTC.RTPTransceiver do
...
+          :udp
+
+        "tcp/" <> _rest ->
+          :tcp
+
+        other ->
+          Logger.debug("Unknown transport protocol: #{inspect(other)}. Assuming UDP")
+          :udp
+      end
+
```
Karolk99: We could add this to the `ExSDP.Utils`. As this case is very SDP-specific, ExSDP seems like the best place to keep it.

What do you think?

Karolk99: Also, I'm not sure about the debug log level. What would happen if the transport protocol was mistaken for the other one?

Karolk99 (commented):
You can add simple tests that check if a transceiver was created with the correct protocol

lib/ex_webrtc/rtp_transceiver.ex
sgfn: this change was reverted

sgfn:
Even though [RFC 4571, sec. 4.](https://www.rfc-editor.org/rfc/rfc4571#section-4) defines a token value `TCP/RTP/AVP` that should seemingly "fit better" in the SDP mlines when using TCP transport, we adhere to the guidelines from [RFC 5764, sec. 8.](https://www.rfc-editor.org/rfc/rfc5764.html#section-8) and always use `UDP/TLS/RTP/SAVPF`. This is because the media stream still uses DTLS and is essentially unchanged, the only difference being handled by ICE (RFC 4571 framing and TCP transport for the would-be UDP datagram payloads).

The same applies to `UDP/DTLS/SCTP` used in data channels.

Karolk99 (approved):

PR #242, opened 2026-01-14
Prevent switching DTLS role

derekkraan:
Hi,

First of all thanks for the great library, this has really been helping us get our use case off the ground.

We started running into an issue where Chrome didn't want to set the remote description, failing with a message like "Failed to set SSL role for the transport".

The full scenario is that the server initializes the initial (voice-only) connection. Then, when the user is ready to share their screen, the user initiates an offer to the server that includes the video track. The server responds with an answer, which Chrome refuses to apply. In the same scenario, Firefox _does_ apply the answer.

Doing some investigation, I came to understand that when there is an existing DTLS connection, Chrome does not tolerate switching roles, while Firefox does. So, even though the RFC mandates `a=setup:actpass` in the offer, Chrome doesn't accept `a=setup:active` in the answer, since in the already-established connection, Chrome is active and the server is passive.

The proposed solution is to track the dtls role, and continue to respond with the same role, as initially established.

In my limited testing, this appears to work with both Chrome and Firefox.

### DIFF
--- a/lib/ex_webrtc/peer_connection.ex
+++ b/lib/ex_webrtc/peer_connection.ex
@@ -645,6 +645,7 @@ defmodule ExWebRTC.PeerConnection do
       dtls_transport: dtls_transport,
       demuxer: %Demuxer{},
       transceivers: [],
+      dtls_role: :new,
       ice_state: :new,
       ice_gathering_state: :new,
       dtls_state: :new,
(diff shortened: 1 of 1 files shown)


sgfn (commented):
Thanks for the fix and the kind words! Seems AOK to me, I'd just ask you to add a few lines of comments in the code to describe why we're doing this -- a shortened version of the commit message will work just fine

derekkraan:
Thanks @sgfn, I've amended the commit to include a [comment](https://github.com/elixir-webrtc/ex_webrtc/pull/242/changes#diff-0b5e8a2cf2a3e6082aeb43b395d89bd6a9eb5a7bee18680c7970275c0702e793R2006-R2008). Is this what you were after?

sgfn:
Yup, that's perfect, thanks!

Now just the `IO.inspect` leftovers:)

derekkraan:
Oops yes of course! Didn't realize I had left those in there. Force-pushed just now to remove those.

sgfn (approved):
🚀

PR #248, opened 2026-03-19
Fix "a struct for ... is expected on struct update" warnings

kidq330:


sgfn:
thanks, closing in favour of #255

PR #251, opened 2026-05-05
Fix crash on un-negotiated DTLS path, delete TURN allocation on socket close

sgfn:
ref: https://github.com/elixir-webrtc/ex_ice/pull/101 https://github.com/elixir-webrtc/ex_turn/pull/10


joaothallis (approved):
Thanks for fixing it! It's working as expected, results of my test: https://github.com/joaothallis/ex_turn_cloudflare_repro/issues/1#issuecomment-4383546516

Karolk99 (approved):

PR #253, opened 2026-05-25
[FCE-2939] Add TWCC send log

sgfn:
Add TWCC send log for outgoing packet departure tracking

- Introduce TWCCSendLog to record departure timestamps and sizes of outgoing RTP packets carrying TWCC sequence numbers, bounded by a 2-second sliding window
- Expose `get_twcc_departure_log/1` public API so users can correlate incoming TWCC feedback with send times for bandwidth estimation
- Rename TWCCRecorder to TWCCRecvLog for clarity, distinguishing receiver-side (arrival) from sender-side (departure) logging

ref: https://github.com/fishjam-cloud/sfu/pull/66

### DIFF
--- a/lib/ex_webrtc/peer_connection/twcc_send_log.ex
+++ b/lib/ex_webrtc/peer_connection/twcc_send_log.ex
@@ -1,0 +1,39 @@
+defmodule ExWebRTC.PeerConnection.TWCCSendLog do
+  @moduledoc false
+  # Sender-side TWCC log.
+  # Records departure timestamps and sizes for outgoing RTP packets
+  # that carry a TWCC sequence number, allowing the end user to correlate
+  # incoming TWCC feedback (TransportFeedback.CC) with send times.
+
+  @packet_window_us 2_000_000
+
+  @type t() :: %__MODULE__{
+          packets: %{non_neg_integer() => {integer(), non_neg_integer()}}
+        }
+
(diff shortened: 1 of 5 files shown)


lib/ex_webrtc/peer_connection/twcc_send_log.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  @spec to_map(t()) :: %{non_neg_integer() => {integer(), non_neg_integer()}}
+  def to_map(%__MODULE__{packets: packets}), do: packets
+
+  defp remove_old_packets(packets, current_time) do
+    min_time = current_time - @packet_window_us
+
+    Map.filter(packets, fn {_seq_no, {departure_time, _size}} ->
+      departure_time >= min_time
+    end)
```
Karolk99: Each time a new packet is recorded, we will go through the whole list, removing probably a single packet. If the list is small enough, that would be okay, but maybe we could use `Map` or `Queue` and optimize the `remove_old_packet` like in the `TWCCRecvLog`.

Karolk99 (approved):

lib/ex_webrtc/peer_connection/twcc_send_log.ex
sgfn: good catch, added a queue

PR #255, opened 2026-05-29
Fix compile-time warnings, update CI

sgfn:
- Fix "a struct for ... is expected on struct update" warnings
- Bump CI to 1.20-otp-29
- Bump credo, dialyzer and fix warnings

Disabled `contract_with_opaque` checks where they failed -- dialyzer keeps yapping about correct code otherwise


Karolk99 (approved):
You can check it with the elixir 1.20

PR #258, opened 2026-07-10
Make get_stats resilient to unresponsive ICE/DTLS transports

roznawsk:
## Type of change
Bug Fix

## Description
`PeerConnection.handle_call(:get_stats)` makes two blocking `GenServer.call`s into child processes — `ice_transport.get_stats` and `DTLSTransport.get_certs_info` (5s default timeout each). If either transport is busy (e.g. a DTLS transport wedged by a packet flood), the call times out and crashes the whole PeerConnection.

This wraps both calls in `try/catch :exit` so a wedged transport degrades the stats output (nil cert info / empty ICE stats) instead of killing the PeerConnection. Late replies after a timeout are absorbed by the existing catch-all `handle_info`.

Fixes FCE-3597 (postmortem FCE-3596: metrics polling crashed a publisher PC, causing ~20 min of WHEP 404s).


sgfn (approved):

lib/ex_webrtc/dtls_transport.ex
```diff
@@ -85,12 +85,12 @@ defmodule ExWebRTC.DTLSTransport do
...
   end
 
-  @spec get_certs_info(dtls_transport()) :: %{
+  @spec get_certs_info(dtls_transport(), timeout()) :: %{
           local_cert_info: cert_info(),
           remote_cert_info: cert_info() | nil
         }
-  def get_certs_info(dtls_transport) do
-    GenServer.call(dtls_transport, :get_certs_info)
+  def get_certs_info(dtls_transport, timeout \\ 5000) do
```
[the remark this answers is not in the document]
roznawsk: bro, grow up

roznawsk:
@sgfn Recap of the latest commit (2d7de2b, "Better timeout handling for `get_stats`"):

The initial crash-proofing had a hole: the two internal transport calls in `handle_call(:get_stats)` used the default 5s `GenServer.call` timeout each, so with both transports wedged the handler could block up to ~10s — while the public `get_stats/1` caller only waits 5s. The caller would still exit with a timeout, i.e. the crash we set out to prevent could still happen one process up.

Changes:
- Both transport calls now use a 1s internal timeout (`@stats_transport_timeout`), so the worst case (~2s) stays under the public 5s budget and `get_stats/1` always returns (degraded) stats instead of raising.
- `DTLSTransport.get_certs_info/2` gained a timeout param; the `ICETransport` behaviour's `get_stats` callback went to arity 2 (note: technically breaking for custom ICE transport implementations, though the module is `@moduledoc false`).
- `get_stats/1` docs now state that an unresponsive transport degrades the corresponding stats to nil/empty values rather than raising.

lib/ex_webrtc/ice_transport.ex
```diff
@@ -57,7 +57,7 @@ defmodule ExWebRTC.DefaultICETransport do
   @impl true
   defdelegate set_remote_credentials(pid, ufrag, pwd), to: ICEAgent
   @impl true
-  defdelegate get_stats(pid), to: ICEAgent
+  def get_stats(pid, timeout), do: GenServer.call(pid, :get_stats, timeout)
```
sgfn: It would be better to change `ExICE.ICEAgent.get_stats` instead

roznawsk: Agreed — done in https://github.com/elixir-webrtc/ex_ice/pull/108 (`ICEAgent.get_stats/2` with a timeout param + 0.16.1 release bump). Once that merges and 0.16.1 is on hex, this line goes back to a `defdelegate` and the dep gets bumped to `~> 0.16.1`.

roznawsk: Agreed, opened PR in ex_ice

sgfn (approved):

